PR #20, opened 2023-05-04
Integrate fcarbon footprint calculation and display

tlestang:
This brings @Llannelongue 's [Green Algorithms](https://www.green-algorithms.org/) footprint calculation to `cats`.

Footprint estimation is optional. It is enabled by specifying the `--jobinfo` option at the command line as follows:

```bash
cats -d 120 --jobinfo cpus=2,gpus=0,partition=CPU_partition,memory=8 
```

Order doesn't matter, but all 4 keywords _must_ be specified. Additionally, job-independant info about the machine (e.g. cpu model) is provided through a YAML config file. A `config.yml` in the current directory is assumed by default, but a specific path can be specified with `--config /relative/path/to/config.yml`.

### Implementation

If `jobinfo` is specified as a cli argument, the job information string is first validated and then parsed into a dictionary (`validate_jobinfo`). 

This dict is then fed to the `carbonFootprint.GreenAlgorithmsCalculator` constructor. All the magic happens in the `carbonFootprint` module, which -- apart from code layout -- is used as-is from @Llannelongue Green Algorithms calculations. Values for carbon footprint are returned through the `get_footprint` method as an `Estimates` (a [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple)) instance. Given the two footprint estimates for running the job now or later, and the potential savings, the `main` functions displays a little report to the user on stdout.

### Notes

- The expected values for `partition` are currently hardcoded but I think they should be derived from the config file.

- The `GreenAlgorotihmsCalculator` constructor expects values for both the average and best intensities at the time `cats` is run. This should be returned along the job start time (probably by `parsedata.writecsv`. Both values are hardcoded for now.

### Reviewers

- I wouldn't make reviewing  carbonFootprint.py a priority for now, since it's essentially a dump of existing code. But it's up to you!


tlestang:
Just made it a draft until I push some updates and give it a more detailed description. Today hopefully

sadielbartholomew:
I'm happy to do a review of this when ready if you'd like? Now we are out of the Hack Day chaos and have time to do formal (GitHub approval etc.) reviewing 😄 !

cats/__init__.py
```diff
@@ -23,17 +26,65 @@ def parse_arguments():
...
+    info = {
+        k: v
+        for k, v in [match.groups() for match in re.finditer(r"(\w+)=(\w+)", jobinfo)]
+    }
+    missing_keys = set(expected_info_keys) - set(info.keys())
+    if missing_keys:
+        print(f"ERROR: Missing job info keys: {missing_keys}")
+        return {}
+    expected_partition_values = ("CPU_partition", "GPU_partition")
+    if info["partition"] not in expected_info_keys:
```
tlestang, line 65: Typo right there: `expected_info_keys` -> `expected_partition_values`

Llannelongue: haha yes was about to comment on this!

Llannelongue:
It looks good, I've just commented out the carbon footprint message, as it is misleading for now (not based on the correct carbon intensities), but good to integrate this code into the main branch to have it in moving forward while we iron out the last details in a new branch

Llannelongue (approved):
:thumbsup:

cats/__init__.py
tlestang: thanks for fixing!

tlestang:
Thanks for the review, I agree with commenting out the hardcoded intensities for now - otherwise it does indeed make it very easy to forget they are :)

Llannelongue:
Thanks for fixing the configgg typo, it was to test the missing file! 

Actually I don't think line 91 should exit? Because if the user doesn't provide a config file but provides a postcode manually (and doesn't ask for carbon footprint), the code can run fine (which is why I initially initialise config with an empty dict)

tlestang:
Ha yes - good point. Sorry I forgot that the config file is actuallly optional. I'm putting your changes back in

Llannelongue:
Are we happy to merge this into main? @tlestang

README.md
```diff
@@ -35,7 +35,49 @@ The postcode is optional, and can be pulled from the `config.yml` file or, if th
...
+    model: "NVIDIA A100-SXM-80GB GPUs" 
+    TDP: 300 # from https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a100/pdf/PB-10577-001_v02.pdf
+    CPU_model: "AMD EPYC 7763" 
+    TDP_CPU: 4.4 # from https://www.amd.com/fr/products/cpu/amd-epyc-7763
+```
+
+Use the `--config` option to specify a path to the config file, relative to the current directory. If no path is specified, `cats` looks for a file named `config.yml` in the current directory.
+
+
+Additionaly, job-specific information must be provided to `cats` through the `--jobinfo` option.  The example below demonstrates running `cats` with footprint estimation for a job using 8GB of memory, 2 CPU cores and no GPU:
```
sadielbartholomew, line 72:
```suggestion
Additionally, job-specific information must be provided to `cats` through the `--jobinfo` option.  The example below demonstrates running `cats` with footprint estimation for a job using 8GB of memory, 2 CPU cores and no GPU:
```

Also is the `jobinfo` CLI option compulsory? Because if it isn't I suggest rewording to avoid the use of 'must' which might imply to people that they have to set that for the command to work. Something like "job-specific information **can** be provided" or similar would be clearer, I think.

README.md
```diff
@@ -35,7 +35,49 @@ The postcode is optional, and can be pulled from the `config.yml` file or, if th
...
+
+```bash
+cats -d 120 --config .config/config.yml \
+  --jobinfo cpus=2,gpus=0,memory=8,partition=CPU_partition
+```
+
+
+>>>>>>> 80eb9f0 (docs: Update readme for carbon footprint estimation)
```
sadielbartholomew, line 80: Looks like this was left over from a merge conflict or similar, so let's blitz it:

```suggestion
```

README.md
Llannelongue: Good point, `--jobinfo` is only compulsory to obtain carbon footprints, otherwise it's not needed indeed

tlestang:
Thank you @sadielbartholomew for taking the time to review this - i'm pressing the green button!

Issue #25, opened 2023-05-09
Implement caching of carbon intensity forecast


tlestang:
Currently a new request to carbonintensity.org.uk is made each time `cats` is run. `In cats/__init__.py`:

```python
def findtime(postcode, duration):
  tuples = get_tuple(postcode) # API request
  result = writecsv(tuples, duration) # write intensity data to disk
				      # as csv timeseries
  # ...
```

Although the carbon intensity data obtained from the API is written on disk, this not taken advantage of. Instead, if the relevant carbon intensity data is already on disk, we'd like to reuse this data instead of making a new request each time.

The local carbon intensity forecast data is reusable if the last forecast datetime is beyond the expected finish datetime of the application, i.e. `forecast_end > now() + runtime`.

A possible approach is to reshuffle the responsabilites of both top-level functions `api_query.get_tuple` and `parsedata.writecsv`.

-   First, `get_tuple` could be responsible for ensuring that the right data is present on disk, and download it if not.
-   Then `writecsv` only cares about computing the best job start time, assuming correct intensity data is available. For instance,

```python
# cats/__init__.py
def findtime(postcode, duration):
  tuples = get_tuple(postcode)
  result = writecsv(tuples, duration)
```

then becomes

```python
# cats/__init__.py
def findtime(postcode, duration):
  # Check if cached carbon intensity data goes beyond
  # now() + duration, download new forecast if not
  # formerly `get_tuple()`
  ensure_cached_intensity_data(postcode, duration)
  # Then -- assuming data is available on disk -- compute
  # the best time to start the job.
  # formerly `writecsv()`
  result = get_best_start_time(duration)
```

This approach has the benefit is maitaining a good separation between talking to the API &#x2013; and caching intensity data &#x2013; and the calculation of the start time. We currently do *almost* have this, expect that the function returning the start time is also responsible for writing the intensity data on disk.

Another possible approach is to push the API query and data caching down to the current `writecsv` function:

```python
def writecsv(data_path: str, duration=None) -> dict[str, int]:
    try:
	return cat_converter(data_path, method, duration)
    except MissingItensityDataError:
	cache_latest_intensity_forecast(postcode)
	return cat_converter(data_path, method, duration)
```

andreww:
If we want, we could cache at the HTTP layer where we call the API. This seems quite easy (see #30).

tlestang:
[...]

PR #30, opened 2023-05-10
Caching of the HTTP call

andreww:
This is a simple way to cache the API request in a way that would be useable on a multi-user system. There are two 'bits'. First, we make the request for the 'top' of the half hour (so hh:00:00 or hh:30:00) by changing the current time before calling the API, then we make the call through requests_cache rather than requests. Because the time (and postcode) is
part of the request this means that old data is
reused, avoiding the HTTP call.

Currently the results are just in file `cats_cache.sqlits` in the current working directory, but we could put this in a hidden file in the users homespace, or in a global temporary directory or whatever.

At this stage think of this as a proof of concept. There are bits that need thinking about (where to put the cache, cleaning up the cache, install the right modules etc.)

### DIFF
--- a/cats/api_query.py
+++ b/cats/api_query.py
@@ -16,9 +16,9 @@ def get_tuple(postcode) -> list[tuple[str, int]]:
     # the current hour or half hour in UTZ
     dt = datetime.now(timezone.utc)
     if dt.minute > 30:
-        dt.replace(minute=30, second=0, microsecond=0)
+        dt = dt.replace(minute=31, second=0, microsecond=0)
     else:
-        dt.replace(minute=0, second=0, microsecond=0)
+        dt = dt.replace(minute=1, second=0, microsecond=0)
     timestamp = dt.strftime("%Y-%m-%dT%H:%MZ")
 
     # Setup a session for the HTTP cache
--- a/cats/api_query.py
+++ b/cats/api_query.py
@@ -22,7 +22,7 @@ def get_tuple(postcode) -> list[tuple[str, int]]:
     timestamp = dt.strftime("%Y-%m-%dT%H:%MZ")
 
     # Setup a session for the HTTP cache
-    session = requests_cache.CachedSession('cats_cache')
+    session = requests_cache.CachedSession('cats_cache', use_temp=True)
     # get the carbon intensity api data
     r = session.get(
         "https://api.carbonintensity.org.uk/regional/intensity/"
--- a/cats/api_query.py
+++ b/cats/api_query.py
@@ -1,5 +1,5 @@
-import requests
-from datetime import datetime
+import requests_cache
+from datetime import datetime, timezone
 from .parsedata import writecsv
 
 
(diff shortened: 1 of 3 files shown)


cats/api_query.py
```diff
@@ -12,9 +12,19 @@ def get_tuple(postcode) -> list[tuple[str, int]]:
     # just get the first part of the postcode
     postcode = postcode.split()[0]
     
+    # get the time (as a datetime object) for the top of
+    # the current hour or half hour in UTZ
+    dt = datetime.now(timezone.utc)
+    if dt.minute > 30:
+        dt.replace(minute=30, second=0, microsecond=0)
```
abhidg, line 19: should be dt = dt.replace(...)

abhidg:
Thanks @andreww, this seems like a good approach. We could put the cache in a temporary folder perhaps, then it can get cleared by the OS. If there is a way to expire the cache after a set time within `requests_cache`, that could be used.

tlestang:
Am I right in understanding that this caches the forecast data for a given 30min time window? For instance, if I make a first request at a time $t$, then any future request within $t$ and $t+30min$ will be cached. But any other request at $t_1 > t_0 + 30min$ will trigger a new API request.

This makes perfect sense to me and is actually I think simpler that what I originally had in mind  as described in #25 .

andreww:
> Am I right in understanding that this caches the forecast data for a given 30min time window? For instance, if I make a first request at a time t, then any future request within t and t+30min will be cached. But any other request at t1>t0+30min will trigger a new API request.

Yes - that's the idea. Except goes back to the start of the previous period. So if the first call is at 17:16 requests will hit the cache until 17:30. This is because it turns out that the API always starts with data from the previous time period.

cats/api_query.py
```diff
@@ -12,9 +12,19 @@ def get_tuple(postcode) -> list[tuple[str, int]]:
     # just get the first part of the postcode
     postcode = postcode.split()[0]
     
+    # get the time (as a datetime object) for the top of
+    # the current hour or half hour in UTZ
+    dt = datetime.now(timezone.utc)
+    if dt.minute > 30:
+        dt = dt.replace(minute=31, second=0, microsecond=0)
+    else:
+        dt = dt.replace(minute=1, second=0, microsecond=0)
```
tlestang, line 21: I'm sorry @andreww I'm still stuck at this! ;) It's now 19:35UTC and a request to the API gives me a forecast with the first timeslot being 19:30 to 20:00. Which -- I think -- is what I want. Replacing the time with 19:31 would lead to exactly the same data, no? 

Same on the other side of the half-hour - if making a requests at 17:16, i'd expect the forecast data to start with `{'from': '17:00', 'to': '17:30'}`.. same for 17:01. What am I missing here?

tlestang: Ah - I guess the `timestamp` must be the same string in the request url for the cache to be hit!

andreww: Exactly - this all assumes the same request URL will always give us the same data (it should, if it's RESTfull) so the code turns the time into a string that represents the half hour where we want the data set to start from. Then it is just a case of asking for the same URL each time we need it within the same half hour window. The other thing to know is that if we were to ask for data starting at 17:00 (not 17:01) first period in the forecast data returned is for 16:30 - 17:00.  Do you think this reasoning needs to go in a comment?

tlestang: All clear now, thank you for clarifying. Yes I think a comment is needed. I would say the assumed knowledge for understanding that bit of code is:

 - That the API yields a list of half-hour time slots from the hour and the half-hour
 - The caching works by making a request to the exact same url (makes sense in retrospect but maybe not completely obvious before hand for people with limited exposure to web services.)
 - Asking for data at HH:00 or HH:30 gives you a forecast starting half and hour too early compared to HH:01 or HH:31.

andreww: I think this is ready for merging.

tlestang: indeed! Thanks for adding the docstrings and comments.

tlestang (approved):

Issue #33, opened 2023-05-11
Documented command to apply with `at` scheduler broken


sadielbartholomew:
HI, a quick one to note that the README command quoted to utilise the output of `cats` with the `at` scheduling command, namely:

> command | at `python -m cats -d <job_duration> --loc <postcode>

is not working. Clearly there is a missing or spurious backtick in there, but eve when one is added in a logical place, namely "\<trivial job command to test\> | at \`python -m cats -d \<job_duration\> --loc \<postcode\>\`" (or I even try it without, just to test to be sure, logically that wouldn't work to my knowledge though), the command won't work, e.g:

```console
$ mkdir mydir | at `python -m cats -d 120 --loc "RG6 6ES"`
{'timestamp': datetime.datetime(2023, 5, 13, 12, 30), 'carbon_intensity': 102.0, 'est_total_carbon': 102.0}
syntax error. Last token seen: B
Garbled time
```

I am not sure of the context, but imagine this command might have worked before the latest changes to include the estimated carbon intensity, because obviously with the output now being a Python-like `dict` it is not a valid `at` timestamp input (unless processed with further commands e.g. by pipe to extract it), whereas if it was just the timestamp before that could have worked assuming the extra backtick.

### Solution for now

Until the CLI input and output format is tidied and we can provide a means to grab the timestamp only, to pass to `at` etc., we could either:
* apply some contrived `awk` pipe command or similar to parse the timestamp out, to get a working command for hooking up to `at`; or
* simply remove the quoted command for now in the README and me say soon it will work such that there is an easy way to hook  up with `at` and other schedulers, etc.

abhidg:
It is not an issue with the dictionary being displayed as that is output to stderr, and at should grab the timestamp directly. There is a `Best job start time` printed that might cause an issue (maybe that is the last token seen 'B' error), but even after removing that, it does not work on my Mac. It appears that macOS and Linux have different default formats for at time specification (@andreww ran into this as well). 

Debian accepts YY-MM-DD (https://manpages.debian.org/bullseye/at/at.1.en.html)

> You can also say what day the job will be run, by giving a date in the form month-name day with an optional year, or giving a date of the form MMDD[CC]YY, MM/DD/[CC]YY, DD.MM.[CC]YY or [CC]YY-MM-DD.

macOS does not (https://ss64.com/osx/at.html)

[...]

Issue #34, opened 2023-05-11
Clarify units and accepted format(s) of `<job_duration>` option value


sadielbartholomew:
I can't find information for this in the README. The code seems to imply the accepted format is an integer representing seconds, but possibly some datetime formats are accepted too (I don't have much time to investigate). Please can someone specify this in more detail in the README.

An example command introduced with a sentence explaining what it does might be really instructive to illustrate the use of both present command-line options (e.g. postcode as a string).

PR #35, opened 2023-05-11
Modify output to be compatible with -t option of at(1)

abhidg:
at(1) time syntax differs between macOS and Linux. Output the time
in MMDDhhmm format that can be passed to the POSIX standard -t option
which works across both.

This also updates the documentation to clarify that job duration
is in minutes and adds an example.

Fixes: #33, #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **C**limate-**A**ware **T**ask **S**cheduler
 
-CATS is a Climate-Aware Task Scheduler. It schedules cluster jobs to minimize predicted carbon intensity of running the process. It was created as part of the [2023 Collaborations Workshop](https://software.ac.uk/cw23). 
+CATS is a Climate-Aware Task Scheduler. It schedules cluster jobs to minimize predicted carbon intensity of running the process. It was created as part of the [2023 Collaborations Workshop](https://software.ac.uk/cw23).
 
 Currently CATS only works in the UK, if you are aware of APIs for realtime grid carbon intensity data in other countries please open an issue and let us know.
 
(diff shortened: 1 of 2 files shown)


colinsauze (approved):

colinsauze (changes requested):
This format isn't giving the correct behaviour on my system (Ubuntu Linux 20.04) when I give the timestamp to at, it interprets 05140300 (03:00 on May 14th) as May 14 17:19:00 2200.

```
$ python -m cats --loc SY23 -d 60
 {'timestamp': datetime.datetime(2023, 5, 14, 3, 0), 'carbon_intensity': 23.0, 'est_total_carbon': 23.0}
 Best job start time: 2023-05-14 03:00:00
05140300
 
$ echo "hello" | at `python -m cats --loc SY23 -d 60`
{'timestamp': datetime.datetime(2023, 5, 14, 3, 0), 'carbon_intensity': 23.0, 'est_total_carbon': 23.0}
Best job start time: 2023-05-14 03:00:00
warning: commands will be executed using /bin/sh
job 18 at Wed May 14 17:19:00 2200
 
$ atq
18      Wed May 14 17:19:00 2200 a colin
```

abhidg:
@colinsauze the output format is meant to be used with the `-t` option of at, try with that and check?

andreww:
On my mac, the longer timestamp works, but does need the `-t` flag:

`echo "hello" | at 202305140300` -> `at: garbled time`
`echo "hello" | at -t 202305140300` -> `job 8 at Sun May 14 03:00:00 2023`

I think we are good to use the longer format (including the year) but our documentation should specify the different invocation of at for these two operating systems.

My guess is we will end up moving this out of the module and into a stand alone python script that gets added to the path and works out what OS it is on somehow.

abhidg:
+1 from me, works on Mac

colinsauze:
I just found why OpenBSD doesn't work, my OpenBSD VM has the time wrong and thinks it is October 20th 2019. But that means it considers May 14th to be in the past. When I set the clock to December 31st it refuses to run a job on January 1st. We definitely want to include the year in our timespec!

andreww (approved):

colinsauze (approved):

Issue #22, opened 2023-05-09
Multinational support: grid APIs for other countries


sadielbartholomew:
From a [Twitter thread](https://twitter.com/RSE_Sam/status/1654141802797047808) relating to `cats`, some folk were enquiring as to whether it works or could work for locations outside of the UK (see original quotes given below for the context, if useful). I agree that it would be nice to provide support for countries not in Britain, assuming of course we can find and use APIs for other national electricity system grids comparable to the National Grid ESO API we have made use of so far.

The first step would be to research whether there are other such APIs we could make use of. Then we can get a feel for how multi-national in scope `cats` could be. Alternatively, we could decide to limit our scope solely to the UK, to avoid the complications. What do people think? It would be especially useful to hear from those with more knowledge of electricity systems than I (I have very little!).

Either way, we should clarify the location scope of `cats` in the documentation. Since our README doesn't mention explicitly that it only works for places in GB, we should add some brief text to clarify that straight away (to be updated if we eventually wider our location scope in line with this Issue). (I'll do that shortly in a commit.)

### Twitter thread background

(See also the link above for original source.)

> Does it work outside the U.S. ? I mean because of the grid data it depends on ?
> > It uses the UK  @NationalGridESO API, but presumably other countries have equivalent ones?
> > > Sorry for assuming U.S. was the default 😉. Probably, there's something equivalent elsewhere, too. Would be cool to add some resources to the http://README.md
> > > > Yeah, people from other countries should definitely raise issues with info for their grid.

colinsauze:
I've added some extra text to the README clarifying it is currently UK only and welcoming links to APIs from other countries.

colinsauze:
One thought is that there is the EPEX Spot website which manages the electricity market for most of Western Europe. They let you get the wholesale price of electricity for the next 24 hours. Price is often a proxy for carbon intensity but doesn't match perfectly.

abhidg:
https://transparency.entsoe.eu/dashboard/show also has breakdown by generation source for Europe, maybe that can be used to predict carbon intensity?

andreww:
What about calculations that have two end-points. For example a data download from the EU?

colinsauze:
We will need to accept location information in a different way to deal with other countries. Not everywhere has postcode. Perhaps we can take a longitude and latitude as well or a city name. We might need to look at geocoding APIs.

AnneHartebrodt:
[...]

PR #37, opened 2023-05-15
Enable extension to other carbon intensity forecast providers

tlestang:
It is currently assumed that carbon intensity data comes from carbonintensity.org.uk. While it is likely to remain the case for a time (until we find data sources for other countries, see #22), it didn't seem a lot of effort -- and changes -- to abstract away the API-specific bits, namely:
- Getting a URL from a postcode and job runtime duration.
- Parsing the api response into a format usable by `cats` -- currently a list of tuples.

This PR add two new arguments `request_url_fun` and `parse_response_fun` to the `parsedata.get_tuple`:

```python
def get_tuple(
    postcode: str,
    request_url_fun: Callable[[datetime, str], str],
    parse_response_fun: Callable[[dict], list[tuple[datetime, int]]],
) -> list[list[tuple[datetime, int]]]:
    """
    get carbon intensity from a web API
    # ...
   """
```

These two arguments are functions and are specified at runtime.

The list of currently supported API interfaces (i.e. the list of avilable implementations of both functions) is provided by a module `api_interface`. This module makes available a dictionary of `APIInterface` instances, which are nothing but namedtuple instances specifying both functions for a given web service.

```python
from cats.apt_interface import API_interfaces

api_interface = API_interfaces["carbonintensity.org.uk"]
tuples = get_tuple(
    postcode,
    api_interface.get_request_url,
    api_interface.parse_reponse_data,
)
```

Isolating the API-specific logic will also help should the carbonintensity.org.uk change its interface in the future. For instance the endpoint syntax or response structure.

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -7,7 +7,7 @@ import yaml
 import sys
 
 from .timeseries_conversion import cat_converter  # noqa: F401
-from .api_query import get_tuple  # noqa: F401
+from .api_query import get_tuple, ciuk_parse_data_from_json, ciuk_request_url  # noqa: F401
 from .parsedata import writecsv  # noqa: F401
 from .carbonFootprint import greenAlgorithmsCalculator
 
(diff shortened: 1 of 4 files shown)


cats/__init__.py
```diff
@@ -7,15 +7,20 @@
 import sys
 
 from .timeseries_conversion import cat_converter  # noqa: F401
-from .api_query import get_tuple  # noqa: F401
+from .api_query import get_tuple
```
andreww, line 10: At some point I expect we'll want to rename get_tuple to something else. But okay for now

andreww (approved):
This looks like a really nice refactor / clean up / preparation for future extension. Let's merge this.

cats/__init__.py
tlestang: Agreed -  I've been trying to resist the urge of making too many changes at once.

PR #38, opened 2023-05-16
Animated gif showing example usage of cats

colinsauze:
This PR adds an animated gif to the readme which shows CATS in action. I used this today in a talk in place of a live demo.


tlestang (approved):
haha - indeed the readme was incomplete until we add at least animated gif :)
Hope the presentation went well!

sadielbartholomew (approved):
Excellent, thanks for creating and adding this, Colin. And +1 for the demo being in a style that would pass all of the Carpentry instructor good practice checklists (no distracting elements, slow enough typing, large easy-to-read text etc.).

We should make sure to update this sufficiently regularly so it doesn't get out-of-date in terms of the CLI and the outputs, as I am sure was the intention.

PR #41, opened 2023-05-27
Pass forecast data without writing csv to disk

tlestang:
Currently the best start time estimate is obtained through calling a hierarchy of functions:

`parsedata.writecsv` -> `timeseries_conversion.cat_converter` -> `timeseries_conversion.get_lowet_carbon_intensity`.

The `writecsv` function writes the forecast data on disk, only for `cat_converter` to read it straight away. I think this was originally meant to be used as a cache, but as it stands it doesn't look to me like writing forecast data to disk is useful to anything else: an API request is made at each execution of `cats`, although since #30  the request itself is cached.

This PR sends forecast data directly to `get_lowest_carbon_intensity`, which is responsible for returning the best job start time and associated average carbon intensity. This reduces complexity by removing the multiple layers between the running `__init__.py` and code that actually optimise the start time.

Instead of plain tuples, forecast data is represented as a list of `CarbonIntensityPointEstimate`.

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -6,7 +6,7 @@ import re
 import yaml
 import sys
 
-from .timeseries_conversion import cat_converter  # noqa: F401
+from .timeseries_conversion import get_lowest_carbon_intensity  # noqa: F401
 from .api_query import get_tuple  # noqa: F401
 from .parsedata import writecsv, avg_carbon_intensity  # noqa: F401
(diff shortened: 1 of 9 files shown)


abhidg (approved):
Looks good, thanks! Can writecsv() be removed now?

tlestang:
I think so - actually the whole `parsedata` module might not be needed anymore?

Issue #42, opened 2023-05-29
How to best estimate an average carbon intensity over the duration of a job


Llannelongue:
We have been discussing with @tlestang about what the data collected by the carbon intensity.org.uk represents, and therefore how best to calculate the average CI over a long period of time. 

The API sends data for 30min periods, each value having a `from` and a `to` parameter (e.g. 50 gCO2e from 7:00am to 7:30am).

There are at least two ways these values can have been obtained (figure below). 

![IMG_3193](https://github.com/GreenScheduler/cats/assets/22586038/5fae5379-7d7c-46ac-a7a4-ec9a4a4dc73d)

Assuming the blue line is the real (continue) CI forecast:
1. The values provided can be timepoints on that curve (red dots) that are then provided for the next 30min (in the example above, the value for 9:00-9:30 would be 40)
2. Or they could have already been averaged over the 30min (green lines). In the example above, the value for 9:00-9:30 would be 45.
3.  The consequence is that in case (1), it is best to approximate the blues curve by using some form of integration between the red dots (trapezoidal was used so far), while in (2), it is best to sum the respective averages (easier). 

Probably best to have both options implemented for when/if we add now APIs, but also good to have a good understanding of what's best.

I'm emailing the people in charge to ask about that, but good to hear everyone's thoughts about that! (as it's quite an important part of the tool!)

tlestang:
carbonintensity.org.uk says (under Methodology):

> The demand and generation by fuel type (gas, coal, wind, nuclear, solar etc.) for each region is forecast several days ahead at 30-min temporal resolution [...]. The forecasts are updated every 30 mins using a nowcasting technique to adjust the forecasts a short period ahead.

The natiogridESO report ([link](https://raw.githubusercontent.com/carbon-intensity/methodology/master/Carbon%20Intensity%20Forecast%20Methodology.pdf)) says:

> The GB carbon intensity $C_t$ at time $t$ is found by weighting the carbon intensity $c_g$ for fuel type $g$ by the generation $P_{g,t}$ of that fuel type. This is then divided by national demand $D_t$ to give the carbon intensity for GB

To me this is indicating that intensity values provided by the API are point estimates rather that the result of an average over the 30min period.

Let's see what they say

tlestang:
@Llannelongue did you hear from the people behind carbonintensity.org.uk about this?

andreww:
Their paper on the methodology says "Estimated carbon intensity data is provided at the end
[...]

Issue #32, opened 2023-05-11
What (should) happen at 48 hrs?


andreww:
I've just done a demo of this and it turned out that the lowest carbon intensity predicted for Oxford was in 48 hrs time (the last half hour returned by the API). Currently we choose to schedule the start of the task then. Is this what we want to happen?

tlestang:
Our first 'live' bug report I think! :champagne: 

> the lowest carbon intensity predicted for Oxford was in 48 hrs time (the last half hour returned by the API). Currently we choose > to schedule the start of the task then. Is this what we want to happen?

No, at least this is not my understanding.

If the returned values of carbon intensities $c_i$ at times $t_i$ with $\delta  = t_\{i+1\} - t_i$. and $\Delta$ is the job runtime provided by the user (using the `-d` flag), then I'm expecting the job start time to be $t_i$ with $i$ minimising the sum
```math
s_i = \sum_{k = i}^{i + M} c_k, 
```
with $M = \lceil \Delta / \delta \rceil$ and $0 \leq i \leq n - M$. So the latest possible job start time is $t_{n-M}$. 

For instance if your job's duration is 3 hours (cats `-d 180`) the latest possible start time should be 45 hours from running `cats`. I thought this is what `timeseries_conversion.get_lowest_carbon_intensity` was doing, but we should have a look.

> Probably worth thinking about some of this kind of edge case, and cooking up some example csv files so we can test them. But > deciding what to do isn't obvious to me.

Couldn't agree more - this stuff is easy to get _almost_ right :)

sadielbartholomew:
(I've nothing helpful to add here right now, but :100: for the use of LaTeX in markdown Thibault, I did not know that was possible!)

colinsauze:
@tlestang's approach reduces the scope for when longer jobs can run but guarantees that you get a sensible answer. Two thoughts on other approaches:

1. Could we add an option to let the user decide.
2. Weather forecasts extend beyond 48 hours with lower accuracy, you can probably make a sensible guess about whether the grid in your region will have a lower carbon intensity up to 5 days from now by looking at the weather forecast. The result won't be as good as the carbonintensity.org.uk predictions but might be good enough to tell you it will be a lot less/more sunny/windy then and your job would be best run a bit later. However this does require us getting into producing our own grid intensity forecast. Although if people are going to wait 4/5 days for a job to run we might need to start asking about how important the job is or asking for a target carbon intensity.

tlestang:
[...]

Issue #26, opened 2023-05-09
Refactor the code for more modularity + self-explanatory function names


Llannelongue:
Looking at the code, I think it would be worth refactoring it to clean up a bit from the hackaday rush and keep each module separately: in particular it would make it easier to address communications between functions, as raised by #25, and facilitate future expansion to new countries by replacing the API part for #22. It would also facilitate asynchronous tasks when some modules need to be bypassed.

An example of what can be confusing at the moment: tracking back where the optimal start time is being computed. In `__init__.py`, the optimal time to run the job is provided by `writecsv` from the `parsedata.py` file (not immediately clear from the name). `writecsv` itself calls `cat_converter` from `timeseries_conversion` which in turns calls the function `get_lowest_carbon_intensity`. The last function makes sense, but the steps in between would be really difficult to guess!

My suggestion is that each component is a separate class in a separate file, for now these are:
- UK API call to obtain carbon intensity time series
- Find the optimal running window to minimise carbon intensities
- Calculate carbon footprint

And each component is called directly from the `main` function, something like: 
```Python
def main(arguments=None):
    parser = parse_arguments()
    args = parser.parse_args(arguments)

    # ... some stuff about config file

    CI = APIcarbonIntensity_UK(args).getCI()

    optimal_time = runtimeOptimiser(args).findRuntime(CI)

    print(f"Best job start time: {optimal_time['start_time']}")

    if carbonFootprint:
        estim = greenAlgorithmsCalculator(...).get_footprint()
        print(f"carbon footprint: {estim}")
```

I'm keen to start working on a branch in this direction, but would like to hear people's thoughts on that as I've probable forgotten some aspects! :smiley:

tlestang:
Sounds good! I agree that a lot of the names are now quite misleading and there is a risk of us getting too comfortable with them ! :) But obviously naming is hard and this was perfectly acceptable in the context of the hack day. I'm thinking that fine-grained changes like renamings can happen progressively as we add missing features and reshuffle the overall design a bit.

Llannelongue:
Yes completely, and actually the naming of individual functions is pretty clear and completely fine taken individually :)  (`writecsv` indeed writes out the carbon intensities), but I think it got a bit muddled when integrating all the modules together last minute. Not a big deal overall

andreww:
[...]

Issue #3, opened 2023-05-04
Calculate corresponding carbon footprints


Llannelongue:

colinsauze:
I think we still need to do some further integration and testing around this one. Did we ever complete the mechanism to get the data from the scheduler to the carbon footprint calculator? Do we still need a way to capture information about job memory and compute requirements?

Llannelongue:
Yes part of the code was last-minute demo-only fixes, so now that we have time it would be good to integrate it neatly. @tlestang did wonders yesterday, so I can liaise with him, and I'll also clean up my code on carbon footprint calculation

tlestang:
hey - #20 is now ready for review. Its a first implementation of footprint calculation and display given machine and job info.

tlestang:
Although one important missing piece is that best and average carbon intensity are still harcoded and should ultimately be returned by `cats` internals alongside the start time (I think?)

tlestang:
Closed by #52

PR #43, opened 2023-05-29
Refactor code (and some other changes)

Llannelongue:
This is following discussions on #26 and during the meeting that now that we are going toward a stable version, it would be worth refactoring the code to make it a bit more future-proof.

This is most likely not final at all, and it'd be great to hear everyone's thoughts to refine it. And also to check that it still works as expected (in particular if I have inadvertently deleted something important!). And we can open dedicated issues for anything arising for discussions here.

Here is a schema of the draft new structure:

cats_refactored drawio

The guiding principles in refactoring were to try to:
- keep each step separate, with a different class/script
- minimised chained imports (file A imports B which imports C, file D also imports B etc.)
- ensure that the main steps can be easily followed from `cats.run`
- make it easy (when possible) to add more API/different optimisation tools in the future

As a reminder, the main steps of CATS are:
1. Read data/information from the command line and the config file 
2. Pull forecast data from a carbon intensity (CI) API
3. Find the window of time (start time + end time) that would minimise average carbon intensity within the known forecast
4. (optional) Estimate the total energy used and carbon footprint by starting or not at the best time 

# An overview of some changes beyond structure

I aimed at keeping the functionalities as they were as much as possible, just trying to enhance when I could. Here are some changed:

- Output as been changed a bit, here is the new one for now:
```
Using config.yml found in current directory
Using carbonintensity.org.uk for carbon intensity forecasts.
Using location provided: PE29

Carbon intensity forecast loaded for between 29/05/2023-20:30 UTC and 31/05/2023-20:30 UTC.

Best start time: 31/05/2023-12:00 UTC (in 39.1 hours)
	 Expected end time: 31/05/2023-17:50 UTC
	 Expected average carbon intensity: 70.03 gCO2e/kWh

Estimated carbon footprint of running job at best time: 10.68 gCO2e
	vs running it now: 13.26 gCO2e (2.58 gCO2e saved)
	vs running at worst time (31/05/2023-01:00 UTC): 17.54 gCO2e (6.87 gCO2e saved)
	- Estimated energy usage: 0.15 kWh
```

- Regarding how to find the best start time:
    - the "simple" method has been removed, as it doesn't really hold to minimise carbon intensity
    - more discussion is needed  regarding what data the carbon intensity.org.uk API sends back. More on this on issue #42. 

- Now all carbon intensity values with timestamps (from the API or computed internally) are stored as a `CarbonIntensityEstimate` (defined in `CI_api_interface.py` but could be defined more centrally). It is a simple adaptation of the previously existing `CarbonIntensityAverageEstimate` and  has properties such as `value`, `start`, `end` and `timedelta` and can be sorted by CI value.

- Added some progress update/error messages throughout (more needed!)

# How this release addresses existing open issues

- #32 : as it stands, it only plans job within the forecast window (latest start time is `end_of_forecast - duration`). How to remove this constraints probably needs to be discussed further as it's not trivial.
-  #22 : this new structure should make it much easier to add new countries, directly in `CI_api_interface.py` (e.g. [here](https://github.com/GreenScheduler/cats/blob/85da24bfe69013605e731d0e69d60a191fd7d5d3/cats/CI_api_interface.py#LL44C1-L44C1))
- It should finish closing #3 as carbon footprint estimates are now fully functional

# Open questions:
1. Why some things that are not errors are being printed on `stderr` and others on `stdout`? As a temporary measure, all updates/messages are sent to `stdout` and only errors/warnings to `stderr`, but would be great to hear some thoughts on that.
2. I suspect I may have broken the pipeline with the `at` linux scheduler with the different output, any thoughts on the best way to pass information on start time?

# What is still needed

Not necessarily before merging, but here is a list non-exhaustive of what still needs to be done (the idea of this PR is to discuss this new structure while polishing the details).
- Write more tests
- Comment the code/functions more thoroughly
- Check how it deals with edge cases, absurd values, etc. (and add error messages)
- Rewrite README
- Check how the new structure works with the `at` linux command


tlestang (commented):
Thanks @Llannelongue  for your work and the detailed PR description. As this is a large set of changes I won't be able to look at everything in one go.

About the API interface:

I don't see the benefit of introducing a new plain class for the API interface, compared to the currently implemented APIInterface namedtuple. It provides the same access to the request url and response parsing functions through dot notation and much simpler implementation.

More problematic to me is that the API-specific logic of forming the query URL and parsing the response is coupled back to the API interface code. Somebody wishing to interface cats with a new web API would have to modify the implementation of `CI_API_interface` itself, adding a new statememt `if self.choice_CI_API == 'myapi.org'` and hardcoding API-specific code in there. I feel like this is a step back from the current implementation (introduced in #37) in which interfacing against a different API is only a matter of providing two functions (possibly grouping them into a new APIInterface object) without having to touch the code of cats itself at any point.

Lastly, do we want to rename `api_query.py` and `api_interface.py` ?

cats/CI_api_interface.py
```diff
@@ -0,0 +1,84 @@
...
+
+    value: needs to be in gCO2e
+
+    """
+    value: float # value needs to be first for the sort function to work
+    start: datetime
+    end: datetime
+
+    def __post_init__(self):
+        self.sort_index = self.value # TODO check whether that's useful
```
tlestang, line 19: It allows to compare two instances of the class. Generally useful if you ask me, but particularly useful in the context of the `WindowedForecast` implementation to be able to write `min(WindowedForecast(data, size))`, see #36

Llannelongue:
Thanks for these @tlestang

> I don't see the benefit of introducing a new plain class for the API interface, compared to the currently implemented APIInterface namedtuple. It provides the same access to the request url and response parsing functions through dot notation and a lot less code.

More on this below after the second comment, but regarding `class` vs `namedtuple`, I don't mind as long as it's easy to switch between APIs (so we would need a master function to pick the right namedtuple in case it's not provided directly by the user). I can draft a new commit reversing to namedtuple.

> More problematic to me is that the API-specific logic of forming the query URL and parsing the response is coupled back to the API interface code. Somebody wishing to interface cats with a new web API would have to modify the implementation of `CI_API_interface` itself, adding a new statememt `if self.choice_CI_API == 'myapi.org'` and hardcoding API-specific code in there. I feel like this is a step back from the current implementation (introduced in #37) in which interfacing against a different API is only a matter of providing two functions (possibly grouping them into a new APIInterface object) without having to touch the code of cats itself at any point.

This is an interesting point, and probably needs more thinking. However not sure about not having to touch the code, I see two different use cases here: 

1. We or other contributors will want to add other CI APIs (e.g. for other countries), and we ideally want to make them part of CATS so that these new APIs are available to the whole community. In this case, it would be good to have all the URL/parsing codes in the same place and `api_inferface` is a good place for that (it also makes it easier to add things by copy-pasting). And in terms of how much hassle it is to add it, it's equivalent now and with the new code (`api_interface` needs to be modified either way, and current code requires messing with `__init__.py` as well), but the existing code doesn't allow user to easily pick an API, this is what the new argument `--api-carbonintensity` introduces.

2. Second use case is if users want to pass their own API wrapper directly to CATS without having to modify the code. And in this case I agree, it would be good to make it possible in an easier way. But how would that work in practice? It would be good to have an idea of how the user would do it if we want to implement it.

I opened issue #44 to have a space to discuss (2) further.

> Lastly, do we really want to rename `api_query.py` and `api_interface.py` ?

 The motivation behind that is to clarify that the code in there is related to pulling carbon intensities, in case more APIs for other parts of the code are included later. But at the moment, this is only future-proofing so I don't mind removing the `CI_` in the names if it's an issue for now.

cats/CI_api_interface.py
Llannelongue: Yes sorting is important, still key to the current implementation where the list of `CarbonIntensityEstimate` is being sorted. However, i don't think that this line is needed to enable the sort, rather what counts is to put `value` in first position. As described in [this StackOverflow post](https://stackoverflow.com/questions/68840767/what-is-the-purpose-of-sort-index-of-a-dataclass-in-python), `self.sort_index` seems to be only needed is the sorting index is not provided when building the dataclass instance (not the case here). Or I may have missed something?

Here is a quick piece of code that tests that:

```python
@dataclass(order=True)
    class valueFirst:
        value: float
        start: datetime
        end: datetime
        def __post_init__(self):
            # self.sort_index = self.value
            self.timedelta = self.end - self.start

foo = valueFirst(start=datetime(2023,5,20,15,30), end=datetime(2023,5,21,15,40), value=18)
bar = valueFirst(start=datetime(2023, 5, 22, 15, 30), end=datetime(2023, 5, 23, 15, 40), value=12)

@dataclass(order=True)
class startFirst_with_index:
    start: datetime
    end: datetime
    value: float
    def __post_init__(self):
        self.sort_index = self.value
        self.timedelta = self.end - self.start

fooo = startFirst_with_index(start=datetime(2023, 5, 20, 15, 30), end=datetime(2023, 5, 21, 15, 40), value=18)
baar = startFirst_with_index(start=datetime(2023, 5, 22, 15, 30), end=datetime(2023, 5, 23, 15, 40), value=12)
```
`foo < bar` returns `False`, as expected (foo.value > bar.value)
But `fooo < baar` returns `True`, because `fooo.start < baar.start` (and the `sort_index` doesn't seem to have an impact)

tlestang: With `order=True` comparison between dataclasses is a tuple comparison (i.e `('b', 2) > ('a', 4)` is `True`. So your first example works because `value` is the first fields and that's what you want the comparison to be based on. Not the case in your second example and yet it works as expected thanks to setting `self.sort_index`.

Llannelongue: But the second example doesn't work as expected, `fooo.value=18` while `baar.value=12`, so the expected behaviour would be `fooo < baar` to be False, which is not the case here.

cats/optimise_starttime.py
```diff
@@ -0,0 +1,91 @@
+from datetime import datetime, timezone
```
tlestang, line 1: I'll add my general comment on this module here so we can have a thread.

I think this PR adds useful missing pieces but I'm a bit uncomfortable with the fast it completely removes the `timeseries_conversion` and `forecast` modules. Makes more sense to me to build upon them instead of discarding them alltogether: they already contain most of the functionality (re)implemented in this module. Particularly, I'm not sure why `WindowedForecast` disappeared -- it was merged 2 days ago -- see #36 . I think it would be straightforward to make it work with a simple summation of intensity values.

That being a few welcome things in this module that are missing in the current implementation:

- Detecting cases where the jon duration exceeds the maximum forecast duration.
- Allowing to compute the windowed/summed/integrated intensity over the window by just summing intervals.
- Handle the fact that the first and last interval in the time window are fraction of the default interval length (30min).

Again -- I think we'd be better off integrating the above in the current code, rather than reimplementing a whole new module from scratch. Unless of course we think that there are strong reasons to do so, but not obvious to me at this stage.

cats/optimise_starttime.py
```diff
@@ -0,0 +1,91 @@
...
+        :param duration: [timedelta]
+        :return: [tuple(CarbonIntensityEstimate, list[CarbonIntensityEstimate])]
+        '''
+        all_averageCIs = self._get_all_averageCIs(duration)
+
+        averageCI_now = all_averageCIs[0]
+        # Find the minimum value
+        all_averageCIs_sorted = sorted(all_averageCIs)
+
+        return all_averageCIs_sorted[0], averageCI_now, all_averageCIs_sorted
```
tlestang, line 91: Here you're sorting a list and selecting the first element. Why not using the builtin `min()` directly?

tests/test_windowed_forecast.py
```diff
@@ -1,45 +1,46 @@
...
-    ]
-    # average
-    expected = [e / (window_size * step) for e in expected]
-
-    assert_allclose(
-        actual=[p.value for p in wf],
-        desired=expected,
-        rtol=0.01
-    )
+# from datetime import datetime, timedelta
```
tlestang, line 1: ?

cats/optimise_starttime.py
Llannelongue: Regarding `WindowedForecast`: should definitely be kept, as highlighted in the PR release note and in the code [here](https://github.com/GreenScheduler/cats/blob/8c927c7c6090c1baa8fc018d802909900c6d9b23/cats/optimise_starttime.py#L38), the goal is to keep the trapezoidal integration as an option, especially while we figure out which one is right for the UK (issue #42). But it's a rather delicate function to get the bounds right (especially first and last segments), so rather than just shoehorn in into this updated structure, it'd be good to discuss about the best way to include it.

Regarding `forecast` it's not removed, just renamed as I'm not sure `forecast` still describes what this part of the code is doing. But it still serves the same purpose. 

As of `timeseries_conversion` (and most of the code in this PR), it's of course reusing existing code, sometimes grouped differently based on how usage has evolved. Which functions in particular from `timeseries_conversion` have been removed completely but still needed? My understanding is that `csv_loader` and `cat_converter` aren't needed anymore (as per PR #41), `check_duration` has just been moved to `check_clean_arguments` as it's what it does and `get_lowest_carbon_intensity` is now in `optimise_starttime` for the same reason.

cats/optimise_starttime.py
Llannelongue: And the worse one a bit later, but yes happy to use min and max instead for now (sorted list may be useful in the future if we need to propose more than one time...but topic for another day.

Llannelongue: Changed in commit [99c88fa](https://github.com/GreenScheduler/cats/pull/43/commits/99c88fafc750066ab1b4c90c6ce6230c70501c1d)

cats/CI_api_interface.py
tlestang: Sorry - I read too fast. I didn't realise `sort_index` is not a special attribute and as you observed, does not have any special meaning on its own. Had to go back reading a bit more on dataclasses  -- it appears to be a trick to impose ordering according to an specific class attribute other than the first one (you probably know this already if you read that SO post linked above).

```python
@dataclass(order=True)
class CarbonIntensityEstimate:
    sort_index: int = field(init = False, repr = False)
    value: float
    start: datetime
    end: datetime

    def __post_init__(self):
        self.sort_index = self.value
```

Or can just make sure that `value` is listed first. But specifying the sort_index makes it explicitly how instances of the class are orderered? Your call.

cats/optimise_starttime.py
tlestang: > the goal is to keep the trapezoidal integration as an option [..]. But it's a rather delicate function to get the bounds right (especially first and last segments), so rather than me just shoehorn in into this updated structure on my own, it'd be good to discuss about the best way to include it.

Don't you think it will more effort re-integrating something that is already there (and covered by a couple of tests) into new code? As opposed to start from what is there. Sorry -- I understand the appeal of writing that module from scratch, but I'm questioning whether it will make the next few steps actually harder. Obviously that's just my own humble opinion, let's see what other people have to say about this!

> Regarding forecast it's not removed, just renamed as I'm not sure forecast still describes what this part of the code is doing. But it still serves the same purpose.

Did `forecast` become `optimise_starttime`? It looks like a new module.

cats/CI_api_interface.py
Llannelongue: I'm thinking of adding a comment to highlight what is being used, especially as `value` is not the only parameter used (in case of equal value, it goes to the next one, and we want to use `start` to start the job as quickly as possible), so the sequence of parameter is important. And that lightens the code. If something like below sounds sensible?

```python
@dataclass(order=True)
class CarbonIntensityEstimate:
    # instances can compared and ordered based on the order of the parameters below.
    # Here, functions such as <, >, min, max will use value first, and in case of equal values, use start
    value: float
    start: datetime
    end: datetime
```

cats/optimise_starttime.py
Llannelongue: I'll submit a commit later today that reintroduces `WindowedForecast` closer to its original form so that we can hopefully get the best of both worlds

Llannelongue: Turns out that properly integrating the first and last time windows into the trapezoidal model was a bit harder than planned (didn't manage with the original implementation but this one is very close) but it's now done with a test: …
[1aca7ab](https://github.com/GreenScheduler/cats/pull/43/commits/1aca7abb0331de809bd9b83f74b48a5fc18a0aa9)

Llannelongue: Also @tlestang if you think testing the integration method on the sinus function is still needed, it can be put back in as it's only commented out for now, for now there is a test on a dummy carbon intensity forecast

Llannelongue:
Are we good to merge that into `main`? From what I can see, it now covers everything already present in `main` (thanks for the feedback! and please do highlight if anything's still missing for users :) ), plus it fixes bugs and edge cases (and hopefully allows more modularity for future-proofing). `windowed_forecast` in particular has been changed compared to first PR here to better build on what the existing module did, and also combine both sum and trapezoidal integration with the right bounds (its exact implementation can of course be revised in another PR if people think it should be coded differently, suggestions much welcome, but at least outputs is accurate for both methods for now)

There are quite a few things still needed, e.g. more extensive tests and checks of arguments for edge cases (but I believe these are still outstanding in `main` anyway, so merging would make it easier to submit future PR)

andreww:
I'm not sure exactly where we are with this one - does it make sense to discuss at the meeting on Wednesday. One possibility would be to try and chop the changes into smaller sets but I think some wider discussion around the architecture could be useful.

My two cents is that a useful approach to thinking about this is to consider other future uses of (maybe bits of) the code base and think about how they would work. So, for cats, I imagine a user who has some big python package providing a service that runs some regular processing, and it's the processing that they want to schedule (no need to integrate with a scheduler). Or, somebody wants to post-process job scheduler logs.

Some (partial) answers to the questions above:

> Why some things that are not errors are being printed on stderr and others on stdout? As a temporary measure, all updates/messages are sent to stdout and only errors/warnings to stderr, but would be great to hear some thoughts on that.

The way the python currently interfaces with `at` is via a shell backtick expansion - stuff that comes out of stdout from cats goes into an argument to `at` and (from the point of view of `at`) everything else is ignored and goes to the terminal. We could sensibly use python's subprocess module to do this instead. This would give us more control but could limit what the user could do.

tlestang:
I'm closing this for now as we discussed it won't be merged as-in. But let's keep the branch around for reference.

PR #45, opened 2023-06-05
Test use of `WindowedForecast`

tlestang:
This adds a couple of tests covering the `WindowsForecast` class. Checking that

- `min(WindowedForecast(data, job_duration))` correctly estimates the average intensity value and bounds of the best runtime period (minimal intensity).
- `WindowedForecast(data, job_duration)[0]` correctly estimates the average intensity value and bounds associated with starting the job at the time of running `cats`, i.e. without waiting for a better time window.


ljcolling (approved):
I think this and #49 can now be merged.

PR #46, opened 2023-06-05
Compute 'now' average intensity estimate alongside optmised one

tlestang:
In order to estimate carbon emission savings, `cats` needs to know:

1. The average carbon intensity associated to running the job at the time of running `cats`
2. The average carbon intensity associated with running the job at the proposed time.

Currently, (1) is computed independently of (2) using a separate function [`parsedata.avg_carbon_intensity`](https://github.com/GreenScheduler/cats/blob/8674f1796aae337ef5bac770f2b7da544f2d271d/cats/parsedata.py#L7). This could however be computed alongside (2) directly as the first item in the `WindowedForecast` object*. 

This makes `get_lowest_carbon_intensity` return two `CarbonIntensityAverageEstimate` values `now` and `then`, corresponding to (1) and (2), respectively. This removes the need for `parsedata.avg_carbon_intensity` and `parsedata` module altogether.

Note that #45 adds a couple of tests checking that (1) and (2) are computed correctly.

-------

*As a reminder, (2) is computed by taking the `min` of a `WindowedForecast` object (see [`get_lowest_carbon_intensity`](https://github.com/GreenScheduler/cats/blob/8674f1796aae337ef5bac770f2b7da544f2d271d/cats/timeseries_conversion.py#L52)), where the window size is the job duration in unit  of forecast time intervals (e.g. 30min).


Llannelongue:
@tlestang I'm a bit confused here as to why continue editing main as PR #43 is still pending? Wouldn't it be best to include these changes into #43, as I think it already addressed some of this? (Test #45 would be great to add in particular)

tlestang:
I'm closing this. It is replaced by #52 which directly onto the new project structure introduced in #49.

PR #50, opened 2023-07-12
Tidy-up command line arguments

Llannelongue:
This is a quick tidy-up of the different command line arguments available and their validation. The goal is to add hints and improve clarity for users, while also being consistent with the current project structure. 

- Drafted "help" text for all arguments (which would show with `-h` menu
- Moved the validation/cleaning of the location to `check_clean_arguments.py` 
- Added an argument to choose a CI API (or in the future, pass their own)
- small clean-up of the `validate_duration` function 
- small clean-up of `validate_jobinfo`
- Print out the values of different arguments (location, config used, CI API used etc.) to `stderr` to confirm values used to users (and simplify debugging for us!)
- Moved datetime 30min standardisation to the interface instead of the query (as this is interface specific, and is linked to the 30min window in the UK) -> I noticed I missed that in the previous PR

### DIFF
--- a/cats/CI_api_interface.py
+++ b/cats/CI_api_interface.py
@@ -5,6 +5,7 @@ from .forecast import CarbonIntensityPointEstimate
 
 
 APIInterface = namedtuple('APIInterface', ['get_request_url', 'parse_response_data'])
+# TODO add a validation function to check the validity of the --location argument
 
 def ciuk_request_url(timestamp: datetime, postcode: str):
     # This transformation is specific to the CI-UK API.
--- a/cats/CI_api_interface.py
+++ b/cats/CI_api_interface.py
@@ -7,9 +7,21 @@ from .forecast import CarbonIntensityPointEstimate
 APIInterface = namedtuple('APIInterface', ['get_request_url', 'parse_response_data'])
 
 def ciuk_request_url(timestamp: datetime, postcode: str):
+    # This transformation is specific to the CI-UK API.
+    # get the time (as a datetime object) and update this to be the 'top' of
+    # the current hour or half hour in UTZ plus one minute. So a call at
+    # 17:47 BST will yield a timestamp of 16:31 UTC. This means that within
+    # any given half hour we will always use the same timestamp.
+    # As this becomes part of the URL, calls can be cached using standard HTTP
+    # caching layer.
+    if timestamp.minute > 30:
+        dt = timestamp.replace(minute=31, second=0, microsecond=0)
+    else:
+        dt = timestamp.replace(minute=1, second=0, microsecond=0)
+
     return (
         "https://api.carbonintensity.org.uk/regional/intensity/"
-        + timestamp.strftime("%Y-%m-%dT%H:%MZ")
+        + dt.strftime("%Y-%m-%dT%H:%MZ")
         + "/fw48h/postcode/"
         + postcode
     )
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -62,15 +62,19 @@ def main(arguments=None):
 
     ## config file
     if args.config:
+        # if path to config file provided, it is used
         with open(args.config, "r") as f:
             config = yaml.safe_load(f)
+        sys.stderr.write(f"Using provided config file: {args.config}\n")
     else:
         # if no path provided, look for `config.yml` in current directory
         try:
             with open("config.yml", "r") as f:
                 config = yaml.safe_load(f)
+            sys.stderr.write("Using config.yml found in current directory\n")
         except FileNotFoundError:
             config = {}
+            sys.stderr.write("WARNING: config file not found\n")
 
     ## CI API choice
     list_CI_APIs = ['carbonintensity.org.uk']
@@ -83,16 +87,20 @@ def main(arguments=None):
 
     if choice_CI_API not in list_CI_APIs:
         raise ValueError(f"{choice_CI_API} is not a valid API choice, it needs to be one of {list_CI_APIs}.")
+    sys.stderr.write(f"Using {choice_CI_API} for carbon intensity forecasts\n")
 
     ## Location
     if args.location:
         location = validate_location(args.location, choice_CI_API)
+        sys.stderr.write(f"Using location provided: {location}\n")
     elif "location" in config.keys():
         location = validate_location(config["location"], choice_CI_API)
+        sys.stderr.write(f"Using location from config file: {location}\n")
     else:
         r = requests.get("https://ipapi.co/json").json()
         postcode = r["postal"]
         location = validate_location(postcode, choice_CI_API)
+        sys.stderr.write(f"WARNING: location not provided. Estimating location from IP address: {location}.\n")
 
     ## Duration
     duration = validate_duration(args.duration)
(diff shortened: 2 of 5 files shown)


cats/CI_api_interface.py
```diff
@@ -7,9 +7,21 @@
...
+    # the current hour or half hour in UTZ plus one minute. So a call at
+    # 17:47 BST will yield a timestamp of 16:31 UTC. This means that within
+    # any given half hour we will always use the same timestamp.
+    # As this becomes part of the URL, calls can be cached using standard HTTP
+    # caching layer.
+    if timestamp.minute > 30:
+        dt = timestamp.replace(minute=31, second=0, microsecond=0)
+    else:
+        dt = timestamp.replace(minute=1, second=0, microsecond=0)
+
```
tlestang, line 21: Yep, good catch!

cats/__init__.py
```diff
@@ -4,21 +4,50 @@
...
+    ### Required
+
+    parser.add_argument("-d", "--duration", type=int, required=True, help="Expected duration of the job in minutes.")
+
+    ### Optional
+
+    parser.add_argument(
+        "--api-carbonintensity", type=str,
+        help="[optional] which API should be used to obtain carbon intensity forecasts. Overrides `config.yml`."
+             "For now, only choice is 'carbonintensity.org.uk' (UK only) (default: 'carbonintensity.org.uk')"
```
tlestang, line 30: Any argument starting with `--` will be considered optional and this should be automatically reflected in the generated help. You can also specify the default value within the call to `add_argument`:
```
parser.add_argument("--api=carbonintensity", type=str, default="carbonintensity.org.uk")
```
that will save you some tests down below.

cats/check_clean_arguments.py
```diff
@@ -20,41 +21,50 @@ def validate_jobinfo(jobinfo: str):
...
+        raise ValueError("--duration needs to be an integer or float (number of minutes)")
+    # make sure it's positive
     if duration_int <= 0:
-        raise ValueError("Windowed method requires timespan to be positive")
+        raise ValueError("--duration needs to be positive (number of minutes)")
 
     return duration_int
+
+def validate_location(location, choice_CI_API):
+    if choice_CI_API == 'carbonintensity.org.uk':
```
tlestang, line 60: Ok for now, but this indicates we need a validation function alongside url and response parsing when defining an API interface

cats/__init__.py
Llannelongue: Good call regarding the default value in general, but here we want the command line to override the config file only if the command line is provided (and default value is only used if neither command line nor config file is specified). If we use a `default` value in `add_argument`, we would lose that information I think? Or would you know a cleaner way to do it?

Llannelongue: You're right that all the `-` or `--` arguments are considered optional by convention, I'll remove the `[optional]` in the description. Looking at the [SO answer here](https://stackoverflow.com/questions/24180527/argparse-required-arguments-listed-under-optional-arguments), we should highlight that `-d` is required (but it shows under "options" for now, so if anyone has recommendations on the best way to deal with it? (there are some suggestions on the SO thread)

cats/__init__.py
```diff
@@ -27,79 +56,103 @@ def main(arguments=None):
...
-            r = requests.get("https://ipapi.co/json").json()
-            loc = r["postal"]
-        else:
-            loc = config["postcode"]
+            config = {}
+            sys.stderr.write("WARNING: config file not found\n")
+
+    ## CI API choice
+    list_CI_APIs = ['carbonintensity.org.uk']
+    if args.api_carbonintensity:
```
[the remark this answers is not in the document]
Llannelongue: Something like this?
```python
    choice_CI_API = 'carbonintensity.org.uk'
    if 'api-carbonintensity' in config.keys():
        choice_CI_API = config["api-carbonintensity"]
    if args.api_carbonintensity:
        choice_CI_API = args.api_carbonintensity
```

cats/check_clean_arguments.py
```diff
@@ -20,41 +21,50 @@ def validate_jobinfo(jobinfo: str):
...
+    info = dict([match.groups() for match in re.finditer(r"(\w+)=(\w+)", jobinfo)])
+
+    # Check if some information is missing
+    if missing_keys := set(expected_info_keys) - set(info.keys()):
+        sys.stderr.write(f"ERROR: Missing job info keys: {missing_keys}")
         return {}
-    expected_partition_values = ("CPU_partition", "GPU_partition")
+
+    # Validate partition value
+    expected_partition_values = config['partitions'].keys()
```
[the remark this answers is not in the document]
Llannelongue: Good call regarding passing only `config['partitions']`.

I'm not sure I follow the second part, `validate_jobinfo` checks the validity of the `--jobinfo` arguments, i.e. `partition`, `memory`, `cpus` and `gpus`?

cats/check_clean_arguments.py
Llannelongue: Indeed good catch, I've added a TODO so we don't forget

cats/__init__.py
tlestang: >  Here we want the command line to override the config file only if the command line is provided [...]. If we use a default value in add_argument, we would lose that information I think?

Hmm! Yes that's tricky. I don't see a neater way now.. Maybe set the default value based on the the content (or non-content) of the config file? Maybe not worth the effort at this point.

tlestang (approved):
Apart from that (minor) [point](https://github.com/GreenScheduler/cats/pull/50#discussion_r1263831447) above, I'm happy for these to go in :)

cats/__init__.py
```diff
@@ -27,79 +56,103 @@ def main(arguments=None):
...
+            sys.stderr.write("\n -!-!- Carbon footprint estimation is a work in progress, coming soon!\n")
+            # Commenting these out while waiting for real carbon intensities
+            # print(f"Estimated emmissions for running job now: {estim.now}")
+            # msg = (
+            #     f"Estimated emmissions for running delayed job: {estim.best})"
+            #     f" (- {estim.savings})"
+            # )
+            # print(msg)
+    else:
+        sys.stderr.write("Not enough information to estimate total carbon footprint, both --jobinfo and config files are needed.\n")
```
sadielbartholomew, line 155: It would be good to define this message earlier under a variable for re-use, since we define the same string twice to use for `stderr` in this block and that way we only need to tweak it once if we want to change it (etc.).

sadielbartholomew (approved):
Looks really good in general. I've raised some minor comments in-line, but once those have been considered I'd also be happy for this to be merged. Thanks Loïc.

cats/__init__.py
```diff
@@ -4,21 +4,50 @@
...
-    parser.add_argument("--jobinfo")
-    parser.add_argument("--config")
+    ### Required
+
+    parser.add_argument("-d", "--duration", type=int, required=True, help="[required] Expected duration of the job in minutes.")
+
+    ### Optional
+
+    parser.add_argument(
+        "--api-carbonintensity", type=str,
```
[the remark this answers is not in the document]
tlestang: Agreed. `--api` and `-a` sounds like a good fit

cats/__init__.py
Llannelongue: Yep good call, done

Llannelongue:
Thanks for the great reviews @tlestang and @sadielbartholomew! All fixed, merging now.

PR #49, opened 2023-07-07
Update project structure

Llannelongue:
The promised PR focusing on project structure, I've copied the figure below from the previous PR (the class names may be obsolete as I haven't changed any functions here but the target structure remains the same). Overall quite light changes as some improvements have now been merged into main already :)

List of changes:
1. Have a dedicated file to check the validity of all the arguments (more checks to be added in the future). Moved validation of `--duration` and `--jobinfo` to this file.
2. The carbon intensity API calls is a separate step as part of `main()`.
3. Starttime optimiser is also a separate step (rather than API and starttime being intertwined)
4. Some other light renaming of functions/files as needed (looking at the files changed, git seems to have lost track of `timeseries_conversion.py`, but it was just renamed `optimise_starttime.py`, same content otherwise).

As usual, thoughts/suggestions welcome! 

image

PR #52, opened 2023-07-14
Re-enable estimation of carbon savings

tlestang:
This PR is an update of #46 following the project structure update of #49.

In order to estimate carbon emission savings, cats needs to know:

- The average carbon intensity associated to running the job at the time of running cats
- The average carbon intensity associated with running the job at the proposed time.

Currently, (1) is computed independently of (2) using a separate function [parsedata.avg_carbon_intensity](https://github.com/GreenScheduler/cats/blob/8674f1796aae337ef5bac770f2b7da544f2d271d/cats/parsedata.py#L7). This could however be computed alongside (2) directly as the first item in the WindowedForecast object*.

This makes `get_avg_estimate` (formerly `get_starttime`) return two CarbonIntensityAverageEstimate values instead of one. The first one is the average carbon intensity corresponding to running the job at the time of running `cats`.  The second one is the best achievable carbon intensity average given the job duration and available data.  As a side effect, this removes the need for the `parsedata` module.


README.md
```diff
@@ -63,6 +63,44 @@ ls | at -t `python -m cats -d 5 --loc OX1`
...
+#### Displaying carbon footprint estimates
+
+`cats` is able to provide an estimate for the carbon footprint
+reduction resulting from delaying your job.  To enable the footprint
+estimation, you must provide information about the machine in the form
+of a YAML configuration file.  An example is given below:
+
+```yaml
+cluster_name: "CW23"
+postcode: "EH8 9BT"
```
Llannelongue, line 75: This should now be `location` instead of postcode

README.md
```diff
@@ -63,6 +63,44 @@ ls | at -t `python -m cats -d 5 --loc OX1`
...
+  GPU_partition:
+    type: GPU
+    model: "NVIDIA A100-SXM-80GB GPUs"
+    TDP: 300
+    CPU_model: "AMD EPYC 7763"
+    TDP_CPU: 4.4
+```
+
+Use the `--config` option to specify a path to the config file,
+relative to the current directory. If no path is specified, `cats`
```
Llannelongue, line 91: I think `--config` can also be an absolute path?

cats/optimise_starttime.py
```diff
@@ -24,7 +28,11 @@ def get_starttime(data, method="simple", duration=None):
...
-        num_intervals = int((duration / interval) + 0.5)  # round to nearest integer (UP)
-        return min(WindowedForecast(data, num_intervals))
+        interval = (
+            data[1].datetime - data[0].datetime
+        ).total_seconds() / 60
+        wf = WindowedForecast(
+            data=data,
+            window_size=ceil(duration / interval)
+        )
+        return wf[0], min(wf)
```
Llannelongue, line 38: As it stands, the function returns one argument if `method=="simple"` and two otherwise, which is not ideal. We could simply remove "simple" as it doesn't make sense from a CO2-minimisation point of view anyway?

cats/__init__.py
```diff
@@ -81,25 +83,20 @@ def main(arguments=None):
...
-        #     f"Estimated emmissions for running delayed job: {estim.best})"
-        #     f" (- {estim.savings})"
-        # )
-        # print(msg)
+        print(f"Estimated emmissions for running job now: {estim.now}")
+        msg = (
+            f"Estimated emmissions for running delayed job: {estim.best})"
+            f" (- {estim.savings})"
+        )
+        print(msg)
```
Llannelongue, line 99: My understanding (but I may have got that wrong) was that for now we output everything to sys.stderr since the `at` command takes stdout as input. We need to clean this up in a future PR to be able to output progress updates in stdout, but for now maybe best to output this to stderr?

Llannelongue:
Looks good, thanks for putting it together @tlestang! I've put small comments, and it should be fine to be merged after that (conflicts with `main` are due to #50 but they look very minor and should be easy to resolve)

README.md
tlestang: Yes it can. Actually that's very standard behavior so no need to specify the type of path..

cats/optimise_starttime.py
tlestang: Agreed, I'd be happy to remove it. But let's adress that in another PR maybe.

cats/__init__.py
andreww: Yes - I think the fix (a different PR for sure) is to wrap the call to `at` inside python rather than having the user use `cats` to generate input to `at` and the command line to pass that through. Doing that portably needs a bit of thought though (and I suspect the idea in my head is not the only sensible design).  See #53

tlestang:
Rebased on `main`

cats/__init__.py
tlestang: Good catch, done.

README.md
tlestang: Done as I rebased this branch on `main`.

config.yml
```diff
@@ -6,9 +6,8 @@
 ## Updated: 12/07/2023
 
 ---
-cluster_name: "CW23"
-location: "EH8"
-api: carbonintensity.org.uk
+location: "EH8 9BT"
```
Llannelongue, line 9: Any reasons why the model `config` file has a full postcode instead of just the first part? This may give the impression to users that this is expected, or that the API uses the whole postcode, while the results are actually only for EH8 here (our code is able to extract the first part, but maybe best to encourage users to only provide it anyway).

Llannelongue (approved):
Looks good!

config.yml
colinsauze: I think it was an example postcode that got used early in the development before we realised we didn't need to specify a full postcode. Go ahead and change it to a shorter postcode.

tlestang: Just did.

tlestang:
Cleaned up the history a bit.
Thanks @Llannelongue and @colinsauze for looking at this.

PR #54, opened 2023-07-23
Account for job start/end time not exactly matching forecast data points

tlestang:
Currently, the average carbon intensity for a job is estimated over a time interval spanning $N + 1$ data points, where $N$ is the `ceil` of the ratio between the job duration and the timestep between data points.  Moreover, cats approximates the job start time as the time of the first available data points. For carbonintensity.org.uk, this the previous top of the hour, or half hour. 

**example**:
current time is 12:48 and my job is 194 minutes long. The job start time will be approximated as 12:30, and the job duration will be approximation as $210min = 7 \times 30min$ where $30min$ is the interval between two data points.

The approach followed in the PR is similar to changes included in #43 , but the implementation fits within the current structure by only enhancing the `WindowedForecast` class. These changes are covered by a new test `test_average_intensity_with_offset`.

**implementation**:
The average carbon intensity over a given time window is estimated by summing the intensity values at the midpoints between each consecutive data points in the window. The first midpoint (`midpt[0]`) is overidden to account for the fact that the first point (used to compute `midpt[0]`) should be located at the job start time. The corresponding intensity value is interpolated between the first and second data point. Similarly, the last midpoint (`midpt[-1]`) is overidden to account for the fast that the last point should be located at the job end time. The corresponding intensity value is interpolated between the penultimate and last data points in the window.

### DIFF
--- a/cats/forecast.py
+++ b/cats/forecast.py
@@ -44,22 +44,22 @@ class WindowedForecast:
             start: datetime,
     ):
         self.data = data
+        self.data_stepsize = data[1].datetime - data[0].datetime
+        self.start = start
+        # TODO: Expect duration as a timedelta directly
         self.end = start + timedelta(minutes=duration)
         self.ndata = bisect_left(data, self.end, key=lambda x: x.datetime) + 1
-        self.start = start
-        self.data_stepsize = data[1].datetime - data[0].datetime
-        # TODO: Expect duration as a timedelta directly
-        self.duration = timedelta(minutes=duration)
 
     def __getitem__(self, index: int) -> CarbonIntensityAverageEstimate:
         """Return the average of timeseries data from index over the
         window size.  Data points are integrated using the trapeziodal
         rule, that is assuming that forecast data points are joined
-        with a straight line. Integral value between two points is the
-        intensity value at the midpoint times the duration between the
-        two points.  This duration is assumed to be unity and the
-        average is computed by dividing the total integral value by
-        the number of intervals.
+        with a straight line.
+
+        Integral value between two points is the intensity value at
+        the midpoint times the duration between the two points.  This
+        duration is assumed to be unity and the average is computed by
+        dividing the total integral value by the number of intervals.
         """
         midpt = [
             0.5 * (a.value + b.value)
@@ -78,7 +78,7 @@ class WindowedForecast:
         i = self.interp(self.data[index], self.data[index + 1], when=start)
         midpt[0] = 0.5 * (i + self.data[index + 1].value)
 
-        end = self.start + index * self.data_stepsize + self.duration
+        end = self.end + index * self.data_stepsize
         i = self.interp(
             self.data[index + self.ndata - 2],
             self.data[index + self.ndata - 1],
--- a/cats/forecast.py
+++ b/cats/forecast.py
@@ -1,3 +1,4 @@
+from bisect import bisect_left
 from math import ceil
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
@@ -45,11 +46,10 @@ class WindowedForecast:
     ):
         self.times = [point.datetime for point in data]
         self.intensities = [point.value for point in data]
-        # Integration window size in number of time intervals covered
-        # by the window.
-        self.data_stepsize = data[1].datetime - data[0].datetime
-        self.window_size = ceil(duration / (self.data_stepsize.total_seconds() / 60))
+        self.end = start + timedelta(minutes=duration)
+        self.ndata = bisect_left(data, self.end, key=lambda x: x.datetime) + 1
         self.start = start
+        self.data_stepsize = data[1].datetime - data[0].datetime
         # TODO: Expect duration as a timedelta directly
         self.duration = timedelta(minutes=duration)
 
@@ -75,8 +75,8 @@ class WindowedForecast:
         v = [  # If you think of a better name, pls help!
             0.5 * (a + b)
             for a, b in zip(
-                    self.intensities[index: index + self.window_size],
-                    self.intensities[index + 1 : index + self.window_size + 1]
+                    self.intensities[index: index + self.ndata - 1],
+                    self.intensities[index + 1 : index + self.ndata]
             )]
 
         start = self.start + index * self.data_stepsize
@@ -85,18 +85,16 @@ class WindowedForecast:
         v[0] = 0.5 * (self.interp(p1, p2, start) + self.intensities[index + 1])
 
         end = self.start + index * self.data_stepsize + self.duration
-        p1 = CarbonIntensityPointEstimate(self.times[index + self.window_size - 1], self.intensities[index + self.window_size - 1])
-        p2 = CarbonIntensityPointEstimate(self.times[index + self.window_size], self.intensities[index + self.window_size])
+        p1 = CarbonIntensityPointEstimate(self.times[index + self.ndata - 2], self.intensities[index + self.ndata - 2])
+        p2 = CarbonIntensityPointEstimate(self.times[index + self.ndata - 1], self.intensities[index + self.ndata - 1])
         v[-1] = 0.5 * (
-            self.intensities[index + self.window_size - 1] + self.interp(p1, p2, end)
+            self.intensities[index + self.ndata - 2] + self.interp(p1, p2, end)
         )
 
         return CarbonIntensityAverageEstimate(
-            start=self.times[index],
-            # Note that `end` points to the _start_ of the last
-            # interval in the window.
-            end=self.times[index + self.window_size],
-            value=sum(v) / self.window_size,
+            start=start,
+            end=end,
+            value=sum(v) / (self.ndata - 1),
         )
 
     def __iter__(self):
@@ -104,4 +102,4 @@ class WindowedForecast:
             yield self.__getitem__(index)
 
     def __len__(self):
-        return len(self.times) - self.window_size - 1
+        return len(self.times) - self.ndata
--- a/cats/forecast.py
+++ b/cats/forecast.py
@@ -1,3 +1,4 @@
+from math import ceil
 from dataclasses import dataclass, field
 from datetime import datetime
 
@@ -36,12 +37,20 @@ class CarbonIntensityAverageEstimate:
 
 class WindowedForecast:
 
-    def __init__(self, data: list[CarbonIntensityPointEstimate], window_size: int):
+    def __init__(
+            self,
+            data: list[CarbonIntensityPointEstimate],
+            duration: int,  # in minutes
+            start: datetime,
+    ):
         self.times = [point.datetime for point in data]
         self.intensities = [point.value for point in data]
         # Integration window size in number of time intervals covered
         # by the window.
-        self.window_size = window_size
+        data_stepsize = (
+            data[1].datetime - data[0].datetime
+        ).total_seconds() / 60
+        self.window_size = ceil(duration / data_stepsize)
 
     def __getitem__(self, index: int) -> CarbonIntensityAverageEstimate:
         """Return the average of timeseries data from index over the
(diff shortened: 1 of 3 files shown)


tlestang:
Looks like this is only working with python3.10+, because I'm relying on being able to pass a sorting key to `bisect`. Probably worth working around this to be able to support 3.9 as well.

andreww (approved):
Changes all look good to me. One question - what happens if somebody submits a task that lasts less than 30 mins? For now, we maybe just want a test case (and I hope it 'just works'). In principle I guess we could start fitting some smooth function to the CI data, but I don't think we want to go there.

cats/forecast.py
```diff
@@ -36,37 +36,97 @@ def __post_init__(self):
...
+        """Return value of carbon intensity at a time between data
+        points, assuming points are joined by a straight line (linear
+        interpolation).
+        """
+        timestep = (p2.datetime - p1.datetime).total_seconds()
+
+        slope = (p2.value - p1.value) / timestep
+        offset = (when - p1.datetime).total_seconds()
+        # import pdb; pdb.set_trace()
+        return p1.value + slope * offset  # Value at t = start
```
tlestang, line 125: Comment is outdated: should be `t = when`

colinsauze:
> Example: You run `cats` at 12:48 for a 6minutes job. The second candidate window is 13:18 to 13:24, located between data points at 13:00 (`data[1]`) and 13:30 (`data[2]`). Both carbon intensity values at 13:18 and 13:24 and interpolated by drawing a straight line between data points at 13:00 and 13:30.
> 
> > I guess we could start fitting some smooth function to the CI data
> 
> I think linear interpolation is enough. If you look at the carbon intensity timeseries, it's already very smooth. In other words the signal doesn't exhibit large variations within 30 min. I guess forecast providers provide data points at an interval ensuring the timeseries is well resolved.

I'd agree that linear interpolation is good enough here. Getting things within the right 30 minute period will achieve what we need in terms of emissions minimisation. 

Part of me is tempted to say the best strategy for sub 30 minute jobs is to randomly determine when they run within the lowest 30 minute window available. This way if many people were launching small jobs they wouldn't all try to run at the same time (even when distributed across multiple unrelated systems).

tlestang:
> Part of me is tempted to say the best strategy for sub 30 minute jobs is to randomly determine when they run within the lowest 30 minute window available. This way if many people were launching small jobs they wouldn't all try to run at the same time (even when distributed across multiple unrelated systems).

Smart. I don't really want to add more features to this PR but that's probably a good starting point for a new one. Or an issue.

colinsauze:
Yes, it's definitely something for a different PR/issue and not critical for getting a minimum viable product ready.

Llannelongue (changes requested):
Thanks for this. I've just ran small test that I had implemented in #43, and for some reason I don't get the expected result. It looks like lbound compute the wrong value here, but it may also well be that I ran the test incorrectly for the current structure. Do you know why it fails?

```python
CI_forecast = [
    CarbonIntensityPointEstimate(datetime=datetime(2023,1,1,8,30), value=26),
    CarbonIntensityPointEstimate(datetime=datetime(2023,1,1,9,0), value=40),
    CarbonIntensityPointEstimate(datetime=datetime(2023,1,1,9,30), value=50),
    CarbonIntensityPointEstimate(datetime=datetime(2023,1,1,10,0), value=60),
    CarbonIntensityPointEstimate(datetime=datetime(2023,1,1,10,30), value=25),
]
wf = WindowedForecast(data=CI_forecast, duration=70, start=datetime(2023,1,1,9,15))
print(wf[0])
start_new = 45
end_new = 60 + (25-60)/30*25
expected_result_trapezoidal = ((start_new+50)*15/2 + (50+60)*30/2 + (60+end_new)*25/2)/70
print(expected_result_trapezoidal)
assert math.isclose(wf[0].value, expected_result_trapezoidal)
```
(I believe the integral between 9:15 and 10:25 should be 49.9.., with the left bound 45 and right bound 30.83..)

cats/forecast.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timedelta
 
 
 @dataclass(order=True)
```
Llannelongue, line 5: Also not a new change, but on the `CarbonIntensityPointEstimate`, I think the `sort_index` introduces unnecessary code. As discussed previously and [here](https://stackoverflow.com/questions/68840767/what-is-the-purpose-of-sort-index-of-a-dataclass-in-python), the first attribute is used for sorting. I'd suggest removing sort_index and just putting a comment there for clarity, something like below. Same thing for `CarbonIntensityAverageEstimate`.

```python
class CarbonIntensityPointEstimate:
    """Represents a single data point within an intensity
    timeseries. Use order=True in order to enable comparison of class
    instance based on the first attribute.  See
    https://peps.python.org/pep-0557

    """
    value: float # the first attribute is used automatically for sorting methods
    datetime: datetime
    
```

cats/forecast.py
```diff
@@ -36,37 +36,97 @@ def __post_init__(self):
...
-        # by the window.
-        self.window_size = window_size
+    def __init__(
+            self,
+            data: list[CarbonIntensityPointEstimate],
+            duration: int,  # in minutes
+            start: datetime,
+    ):
+        self.data = data
+        self.data_stepsize = data[1].datetime - data[0].datetime
```
Llannelongue, line 46: This assumes that all the intervals are the same in the forecast, which may be a constraint down the line. And is it right that `bisect_left` below is there to avoid making this assumption? (otherwise, `ndata` could be calculated as `(end - start)/stepsize` or something like that).
It could be good to pick one (and if we assume all intervals are the same for now for simplicity, maybe worth leaving a TODO or comment to get back to it later).

cats/forecast.py
```diff
@@ -36,37 +36,97 @@ def __post_init__(self):
...
-                    self.intensities[index + 1 : index + self.window_size + 1]
-            )]
 
+        # Account for the fact that the start and end of each window
+        # might not fall exactly on data points.  The starting
+        # intensity is interpolated between the first (index) and
+        # second data point (index + 1) in the window.  The ending
+        # intensity value is interpolated between the last and
+        # penultimate data points in he window.
+        window_start = self.start + index * self.data_stepsize
```
Llannelongue, line 82: One potential downside of doing all the calculations in the `__getitem__` method is that it requires redoing calculations every time (I think?): as in, `min` already calculates all windows, but then `w[0]` does it again etc.. An alternative would be to calculate all windows once, store it in self and only access it after that. Or are there performance benefits to doing everything here?

cats/forecast.py
tlestang: > This assumes that all the intervals are the same in the forecast

Yes

> it right that bisect_left below is there to avoid making this assumption? (otherwise, ndata could be calculated as (end - start)/stepsize or something like that).

This is how I originally implemented it, but here is a counter example. Say you've got 3 data points at 12:30, 13:00 and 13:30. Your job is 10 minutes long.  If you start at 12:40, you end at 12:50: that's `ndata =2`.  But if start time and end time fall on both side of a data point, say start at 12:55 and end at 13:05, then `ndata = 3`. Hope this helps? Sorry I wish we could draw on a board :)

cats/forecast.py
tlestang: Overall I think performance should not be a concern here. Even then, only the the first window is computed twice. But yes, if you were to run `min` twice, you'd compute all windows twice.

cats/forecast.py
tlestang: Agreed! I won't add that to that PR, but feel free to make that change of you happen to have some time for it

tlestang:
Thanks for testing @Llannelongue because I think you've exposed a mistake! In __getitem__

```
        return CarbonIntensityAverageEstimate(
            start=window_start,
            end=window_end,
            value=sum(midpt) / (self.ndata - 1),
        )
```

dividing by `self.ndata - 1` assumes all midpoints are regularly spaced: true for inner points but not the first and last one, based on interpolated data points. I've also made this assumption in the tests I believe... :face_with_head_bandage:

tlestang:
I changed the average calculation to account for the difference in weights. @Llannelongue I added a test case based on your test above. 

> I believe the integral between 9:15 and 10:25 should be 49.9.., with the left bound 45 and right bound 30.83..)

Agreed. I think it is now ;)

cats/forecast.py
Llannelongue: Ok, I've just added a a note #57 to remember to either fully check that or make it work for any interval sizes

tests/test_windowed_forecast.py
```diff
@@ -107,3 +111,114 @@ def test_average_intensity_now():
...
+            0.5 * (45 + 50) * 15 +
+            0.5 * (50 + 60) * 30 +
+            0.5 * (60 + 30.83) * 25
+        ) / duration
+    )
+    assert result.start == expected.start
+    assert result.end == result.end
+    # Allow for 1% relative tolerance between result and expected
+    # value.  Difference is probably due to truncating the
+    # second interpolated value above.
```
Llannelongue, line 147: Thanks for fixing that! `result` in this function is getting the right result, but  `expected` is a bit off, wouldn't it be better to fix `expected` rather than raising the error tolerance of the test? Simple way: `60 + (25-60)/30*25` instead fof `30.83`.

tests/test_windowed_forecast.py
```diff
@@ -107,3 +111,114 @@ def test_average_intensity_now():
...
+        CarbonIntensityPointEstimate(datetime(2023,1,1,9,0), value=40),
+        CarbonIntensityPointEstimate(datetime(2023,1,1,9,30), value=50),
+        CarbonIntensityPointEstimate(datetime(2023,1,1,10,0), value=60),
+        CarbonIntensityPointEstimate(datetime(2023,1,1,10,30), value=25),
+    ]
+    duration = 70  # in minutes
+    # First available data point is for 08:00 but the job
+    # starts 15 minutes later.
+    job_start = datetime.fromisoformat("2023-01-01T08:45")
+    result = WindowedForecast(CI_forecast, duration, start=job_start)[1]
```
Llannelongue, line 132: It might be me misunderstanding the role of `start` in `WindowedForecast`: how come the two rows below don't yield the same values despite being the same time windows?
```
result = WindowedForecast(CI_forecast, duration, start=datetime.fromisoformat("2023-01-01T08:45"))[1]
>>> CarbonIntensityAverageEstimate(start=datetime.datetime(2023, 1, 1, 9, 15), end=datetime.datetime(2023, 1, 1, 10, 25), value=49.970238095238095)

result2 = WindowedForecast(CI_forecast, duration, start=datetime.fromisoformat("2023-01-01T09:15"))[0]
>>> CarbonIntensityAverageEstimate(start=datetime.datetime(2023, 1, 1, 9, 15), end=datetime.datetime(2023, 1, 1, 10, 25), value=49.75595238095238)
```

cats/forecast.py
Llannelongue: Addressed with #58

tests/test_windowed_forecast.py
tlestang: Another good catch: that's because the code in `__getitem__` assumes that the job start time (`09:15` in the above) is always in between the first and second data points. I think that's always true in the context of cats since data from the API always starts at the first available point _before_ the time of the request.

tlestang: But I realise now that it might be easy to get rid of this assumption: just trim the `data` passed to `WindowedForecast.__init__` removing all data points before the passed `start` time, except the point directly before. I'll give that a try later.

tlestang: I just made that change. If I'm correct it should work as you (reasonably) expected above.

Llannelongue (approved):

tlestang:
Thank you @andreww @colinsauze and @Llannelongue for reviewing this. It really helped make this better!

PR #55, opened 2023-07-26
Remove `simple` method for getting the job start time.

tlestang:
This removes the so-called `simple` method of getting the start time, which only returns the minimum point estimate across the forecast data. This came about [while reviewing #52](https://github.com/GreenScheduler/cats/pull/52#discussion_r1268641032). If I remember correctly, the `simple` method is inherited from the race of the CW23 hackathon, as a good way of having something quick to interface with the other modules in cats.

This also remove `test_timeseries_conversion`, which I think is made obsolete by tests in `test_forecast`.


abhidg (approved):
Thanks!

Llannelongue (approved):

PR #60, opened 2023-08-07
docs: cli reference

ljcolling:
added cli arguments to docs


tlestang:
Thanks, don't you need to `pip install cats` in the [workflow](https://github.com/GreenScheduler/cats/blob/main/.github/workflows/docs-build-deploy.yml)? In case autodocs needs to `import cats.__init__`. I can see that's one of the reason the API reference isn't being generated by sphinx (plus the fact that module names are outdated after merging #49)

ljcolling:
> Thanks, don't you need to `pip install cats`

Ah yes you do. I'll update it

ljcolling:
Ok, I think it should work as expected now

tlestang:
Yes, I pushed your changes to my fork and built it from there: https://tlestang.github.io/cats/api-reference.html

docs/source/api-reference.rst
```diff
@@ -47,25 +47,17 @@ Modules
...
+``cats.optimise_starttime``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. automodule:: cats.timeseries_conversion
+.. automodule:: cats.optimise_starttime
     :members:
 
-
 Python objects
 --------------
```
tlestang, line 60: Do we need [this section](https://tlestang.github.io/cats/api-reference.html#python-objects)? It seems redundant with the module documentation above.

sadielbartholomew: Good point Thibault, we can get rid of it. Can you remove the section here in this PR, please Lincoln?

sadielbartholomew (approved):
Confirmed to fix the issue at hand. Excellent, thanks Lincoln.

PR #62, opened 2023-09-12
Minor documentation updates for consistency and clarity

sadielbartholomew:
Quick one to make some minor formatting updates, mostly for the command-line help descriptions (I went in to update the documentation README and thought I should include these updates), namely:

* instead of listing the documentation build requirements in the README, pointing to the new `docs-requirements.txt` file created in #60 so we don't have two equivalent lists to keep in sync;
* being more consistent in the command-line descriptions by, e.g., using only single quotes around all keywords (means the generated CLI reference emphasises them via italics but avoiding double quotes which will clutter the terminal `--help` output) and listing defaults at the end;
* adding default information for `--jobinfo` which was missing before, to be explicit and clear;
* capitalising the first word from one docstring which was inconsistent with the others coming through in the API reference.

### DIFF
--- a/cats/CI_api_query.py
+++ b/cats/CI_api_query.py
@@ -5,7 +5,7 @@ from .forecast import CarbonIntensityPointEstimate
 
 def get_CI_forecast(location: str, CI_API_interface) -> list[CarbonIntensityPointEstimate]:
     """
-    get carbon intensity from an API
+    Get carbon intensity from an API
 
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
(Pretty trivial PR to review, so open to anyone as reviewer. Thanks!)

andreww (approved):
Looks good to me.

PR #75, opened 2024-03-09
Add pre-commit configuration and autoformat

### LINKED ISSUE #66: Configure pre-commit config
Suggested tool `black`, `isort` in #29, could also use `ruff`.

abhidg:
This PR adds pre-commit configuration. I have opted to use [`ruff`](https://docs.astral.sh/ruff/) as the linter, open to alternatives. Reason for choosing `ruff` over `flake8`, `black` is that it is one tool that combines most Python autoformatting and linting functions in one and is very fast.

- Add pre-commit configuration, fixes #66
- Initial pre-commit run with ruff autoformatting
- Add git blame ignore revisions file


.vscode/settings.json
tlestang, line 1: Do we need vs code specific config in the git repo?

tlestang (approved):
Ah you beat me to it!! I also went with ruff, so that's a yes for me :) thanks @abhidg

.vscode/settings.json
abhidg: Don't think so - removing!

ljcolling:
Good choice with Ruff, and it kinda fits with the ethos of the project

PR #77, opened 2024-03-10
Move processing of cli args and config file to separate module

tlestang:
The configuration for cats is specified by both a config file and cli arguments. Values provided at the command line often
override the values provided in the config file. It means there is a bit of work to be done in order to distil all that information into a fixed runtime configuration before cats can start making a request to a CI service. Namely, cats needs
- The location
- The APIInterface instance that prescribes how to contact the CI service
- The job duration

This PR moves the processing required to set these three things (the cats runtime configuration) into a separate module `configure` that provides a single `configure` function as an entry point. Benefits are:

- Less code in `__init__.py` which is starting get big
- Easier to add unit tests around the processing of config file and cli arguments

### DIFF
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -11,6 +11,7 @@ from collections.abc import Mapping
 from typing import Any
 
 import yaml
+import requests
 
 from .CI_api_interface import API_interfaces, APIInterface
 
@@ -18,8 +19,9 @@ from .CI_api_interface import API_interfaces, APIInterface
 def configure(args):
     configmapping = config_from_file(configpath=args.config)
     CI_API_interface = CI_API_from_config_or_args(args, configmapping)
+    location = get_location_from_config_or_args(args, configmapping)
 
-    return configmapping, CI_API_interface
+    return configmapping, CI_API_interface, location
 
 
 def config_from_file(configpath = "") -> Mapping[str, Any]:
@@ -55,3 +57,19 @@ def CI_API_from_config_or_args(args, config) -> APIInterface:
             f"Error: {api} is not a valid API choice. It must be one of "
             "\n".join(API_interfaces.keys())
         )
+
+
+def get_location_from_config_or_args(args, config) -> str:
+    if args.location:
+        logging.info(f"Using location provided: {location}")
+        return args.location
+    if "location" in config.keys():
+        logging.info(f"Using location from config file: {location}")
+        return config["location"]
+
+    logging.warning(
+        "location not provided. Estimating location from IP address: "
+        f"{location}."
+    )
+    r = requests.get("https://ipapi.co/json").json()
+    return r["postal"]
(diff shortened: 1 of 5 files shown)


abhidg (commented):
Looks good! I would mock out the `ipapi.co` endpoint to make the tests reproducible, using something like https://github.com/jamielennox/requests-mock or Python's own mock

tlestang:
Thanks for the pointers, I never had a chance to use a mock before. It does look a lot better.

cats/configure.py
```diff
@@ -0,0 +1,116 @@
...
+            return {}
+
+
+def CI_API_from_config_or_args(args, config) -> APIInterface:
+    try:
+        api = args.api if args.api else config["api"]
+    except KeyError:
+        api = "carbonintensity.org.uk"  # default value
+        logging.warning(
+            "Unspecified carbon intensity forecast service, " f"using {api}"
```
sadielbartholomew, line 81:
```suggestion
            f"Unspecified carbon intensity forecast service, using {api}"
```

cats/configure.py
```diff
@@ -0,0 +1,116 @@
...
+        logging.error(
+            "Could not get location from ipapi.co.\n"
+            f"Got Error {r.status_code} - {r.json()['reason']}\n"
+            f"{r.json()['message']}"
+        )
+        sys.exit(1)
+    location = r.json()["postal"]
+    assert location
+    logging.warning(
+        "location not provided. Estimating location from IP address: " f"{location}."
```
sadielbartholomew, line 114:
```suggestion
        f"location not provided. Estimating location from IP address: {location}."
```

docs/source/api-reference.rst
```diff
@@ -23,6 +23,12 @@ Modules
 .. automodule:: cats.__main__
     :members:
 
+``cats.configure``
+^^^^^^^^^^^^^^^^^^^
```
sadielbartholomew, line 27:
```suggestion
^^^^^^^^^^^^^^^^^^
```

to match the character length of the title, else Sphinx will raise a warning.

cats/configure.py
```diff
@@ -0,0 +1,116 @@
...
+            f"Error: {api} is not a valid API choice. It must be one of " "\n".join(
+                API_interfaces.keys()
+            )
+        )
+
+
+def get_location_from_config_or_args(args, config) -> str:
+    if args.location:
+        location = args.location
+        logging.info(f"Using location provided: {location}")
```
sadielbartholomew, line 96:
```suggestion
        logging.info(f"Using location provided from command: {location}")
```

to be more specific perhaps, since 'provided' could also include provision by configuration file?

sadielbartholomew (approved):
Very good idea to move the CLI processing out. Looks all good to me, though whilst here I have raised some minor comments related to formatting (mostly from `ruff` or similar condensing down string lines, to tidy). And I agree with @abhidg that some use of `mock` would be ideal, but happy for you to merge once you've considered those.

cats/configure.py
tlestang: Well done foe spotting that!

tlestang:
This is going in, thanks @abhidg and @sadielbartholomew for the review!

PR #78, opened 2024-03-15
Increase coverage, at 80% now

abhidg:
- Add type hints, stricter checks in validate_jobinfo()
- Add test for validate_jobinfo()
- Increase API query test coverage
- Add tests for output formatting

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -154,8 +154,8 @@ class CATSOutput:
 
         if self.emmissionEstimate:
             out += (
-                f"Estimated emmissions for running job now: {self.emmissionEstimate.now}\n"
-                f"Estimated emmissions for running delayed job: {self.emmissionEstimate.best})\n"
+                f"\nEstimated emmissions for running job now: {self.emmissionEstimate.now}\n"
+                f"Estimated emmissions for running delayed job: {self.emmissionEstimate.best}"
                 f" (- {self.emmissionEstimate.savings})"
             )
         return out
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -186,7 +186,7 @@ def schedule_at(output: CATSOutput, args: list[str]) -> None:
     )
 
 
-def main(arguments=None):
+def main(arguments=None) -> Optional[int]:
     parser = parse_arguments()
     args = parser.parse_args(arguments)
     if args.command and not args.scheduler:
@@ -194,7 +194,7 @@ def main(arguments=None):
             "cats: To run a command with the -c or --command option, you must\n"
             "      specify the scheduler with the -s or --scheduler option"
         )
-        sys.exit(1)
+        return 1
     config, CI_API_interface, location, duration = get_runtime_config(args)
 
     ########################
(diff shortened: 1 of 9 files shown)


cats/__init__.py
```diff
@@ -255,4 +255,4 @@ def main(arguments=None):
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
```
tlestang, line 258: Are there practical benefits to always returning from `main` instead of directly calling `sys.exit` from there?

cats/check_clean_arguments.py
```diff
@@ -22,26 +28,29 @@ def validate_jobinfo(jobinfo: str, expected_partition_names):
         "cpus",
         "gpus",
     )
-    info = dict([match.groups() for match in re.finditer(r"(\w+)=(\w+)", jobinfo)])
+    info = dict([match.groups() for match in re.finditer(r"(\w+)=([\w.]+)", jobinfo)])
```
tlestang, line 31: Was the previous regexp missing something?

tests/test_main.py
```diff
@@ -0,0 +1,47 @@
...
+    "GBR",
+)
+
+
+@pytest.mark.skip
+def test_schedule_at():
+    schedule_at(OUTPUT, ["ls"])
+    assert now_start.strftime(AT_OUTPUT) in subprocess.check_output(["atq"]).decode(
+        "utf-8"
+    )
```
tlestang, line 31: Is there a point to skipping this test? I guess it'll fail because the command is `atq` and not `at`

cats/__init__.py
abhidg: Makes it easier to test for failure conditions by checking the return code, alternatively could use pytest.raises(SystemExit) to test this.

cats/check_clean_arguments.py
abhidg: Yes, it would silently drop fractional parts, so `cpus=2.5` would be parsed as `{ "cpus": "2" }`. We do not have any floats in the jobinfo string, but may in the future.

tests/test_main.py
abhidg: Forgot to disable the skip! This is checking that the job is actually scheduled -- atq lists the at queue.

abhidg: at seems to be unavailable in the GitHub CI env, so will rework to use a mock

tlestang:
Sorry I didnt notice this was in draft mode!

cats/check_clean_arguments.py
tlestang: Nice!

tlestang:
The `validate_jobinfo` is removed in #80 so this is going to clash with your changes here.  Do you think it's worth adding similar typing in [`configure.get_job_info`](https://github.com/GreenScheduler/cats/blob/571709660162b53202e79621d59f9a716163240f/cats/configure.py#L140) ?

tlestang (approved):

PR #80, opened 2024-03-18
Reduce carbon footprint estimation to total energy amount

tlestang:
This PR trims down the carbon footprint estimation module [carbonFootprint.py](https://github.com/GreenScheduler/cats/blob/main/cats/carbonFootprint.py). 

For the purpose of providing an estimation of carbon footprint savings, `cats` only needs to know the total carbon footprint, i.e. summed across all devices (gpus, cpus) and memory. This is straightforward to compute given the new `jobinfo` format introduced in #79:

```python
energy = (
    PUE
    * (runtime.total_seconds() / 3600)
    * sum([(nunits * power) for nunits, power in jobinfo])
    / 1000
    )
best = energy * average_best_ci
now = energy * average_now_ci
```

where `energy` is the *total* energy. I deliberately removed code accounting for energy for a specific device type, since this is not used in cats so far.

This also removes `greenAlgorithmsCalculator` methods formatting an amount of gCO2 into units of driving, flying...etc. This is probably a nice to have, but AFAIK this is dead code. Anyone feel free to add them back if they wish to make use of them.

Finally, this gets rid of the `check_clean_arguments.validate_job_info` function since the job information required for the carbon estimation footprint is fully specified by the `jobinfo` list returned by `configure.get_runtime_configuration`.


cats/__init__.py
```diff
@@ -219,7 +218,7 @@ def main(arguments=None):
             "      specify the scheduler with the -s or --scheduler option"
         )
         sys.exit(1)
-    config, CI_API_interface, location, duration = get_runtime_config(args)
+    CI_API_interface, location, duration, jobinfo, PUE = get_runtime_config(args)
```
tlestang, line 221: The `config` dict representing the yaml config file can now be confined to the `configure` module and not appear in the `__init__` module. The `jobinfo` list of `(ndevice, power)` is all that is require for carbon footprint estimation down the line.

abhidg (approved):
Looks good to me!

PR #82, opened 2024-04-08
Fix test failures on main

andreww:
This fixes two (of the three) failing tests on main.

The issues that are fixed involves a test case where newlines are printed by cats, which are not in the reference output in the tests (an easy fix).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.10
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
@@ -39,6 +39,11 @@ jobs:
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Check types with mypy
+      run: |
+        python3 -m pip install mypy
+        python3 -m pip install types-PyYAML types-redis types-requests types-ujson
+        python3 -m mypy cats
     - name: Test with pytest
       run: |
         python3 -m pytest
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -39,6 +39,11 @@ jobs:
         flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Check types with mypy
+      run: |
+        python3 -m pip install mypy
```
abhidg, line 44: Ideally these should be in test dependencies and be installable in one line

andreww: You mean inside `project.optional-dependencies` in pyproject.toml?

andreww: I see - done this now.

andreww:
I think some of my "fixes" in here are properly addressed in #80, 058a780 in particular. @tlestang - we should probably figure out how to order these two PRs, or if it's best to chop this one up / just take the stuff that's not involved in #80.

tlestang:
Hi @andreww  - I split the rewriting/testing of the carbon footprint estimation module into two PR #79 and #80. I just merged #80 in which fixed the failing test on `main`. I also resolved a few inconsistencies (unused function or variable like `jobinfo`) that were necessary to introduce in order to integrate the changes in two distinct steps without breaking `cats`.

cats/configure.py
```diff
@@ -24,7 +24,8 @@
 __all__ = ["get_runtime_config"]
 
 
-def get_runtime_config(args) -> tuple[dict, APIInterface, str, int]:
+def get_runtime_config(args) -> tuple[Mapping[str, Any], APIInterface, str, int, 
+                                      Union[list[tuple[int, float]],None],Any]:
```
tlestang, line 28: one of these cases where typing really isn't worth the trouble in my opinion.

tlestang:
About the conflicts:
- The `check_clean_arguments.py` was removed, because the job of processing and validating command line arguments is now carried out in the config module.
- After #80 the in-memory representation of the configuration file (`config` dictionary) doesn't have to leave the config module, so the `config.get_runtime_config` function doesn't have to return it in `__init__.py`. Same goes for `_jobinfo`.

andreww:
Thanks - I'll reorder and rebase this branch so that it makes sense now #80 is merged.

andreww:
Rebased and cleaned all of this up. Two new PRs inbound. We can kill this one.

PR #84, opened 2024-04-17
Turn on static type checks in the CI (and have them pass)

andreww:
A reasonable amount to this, but it only (a) enabled type checking in our CI workflow and (b) makes this pass. It's a rebase of a previous PR with commits reorganised.

With the changes to get_runtime_config's return types, I think I agree with @tlestang that this is one of those places where type checking isn't really worth it. But if we don't do anything about it we cannot really do type checking elsewhere. I think (as I say in fb01f50dcc327b7c830b4cf07142dff16a76cc2d) that fact that we check `args.footprint` in two places isn't ideal (what if we rename the argument) but I wouldn't worry too much about that unless we actually want to further refactor. Anyway, I kept all of that in a single commit.

The other changes are (I think) pretty straightforward. I've incorporated the changes suggested by @abhidg in the previous PR.

### DIFF
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -13,7 +13,7 @@ runtime configuration consits of:
 import logging
 import sys
 from collections.abc import Mapping
-from typing import Any
+from typing import Optional, Any
 
 import requests
 import yaml
@@ -24,7 +24,9 @@ from .constants import MEMORY_POWER_PER_GB
 __all__ = ["get_runtime_config"]
 
 
-def get_runtime_config(args) -> tuple[dict, APIInterface, str, int]:
+def get_runtime_config(args) -> tuple[APIInterface, str, int, 
+                                      Optional[list[tuple[int, float]]],
+                                      Optional[float]]:
     """Return the runtime cats configuration from list of command line
     arguments and content of configuration file.
 
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -47,7 +47,7 @@ def get_runtime_config(args) -> tuple[dict, APIInterface, str, int]:
     try:
         duration = int(args.duration)
     except ValueError:
-        logging.eror(msg)
+        logging.error(msg)
         raise ValueError
     if duration <= 0:
         logging.error(msg)
@@ -91,13 +91,14 @@ def CI_API_from_config_or_args(args, config) -> APIInterface:
         api = "carbonintensity.org.uk"  # default value
         logging.warning(f"Unspecified carbon intensity forecast service, using {api}")
     try:
-        return API_interfaces[api]
+        interface = API_interfaces[api]
     except KeyError:
         logging.error(
             f"Error: {api} is not a valid API choice. It must be one of " "\n".join(
                 API_interfaces.keys()
             )
         )
+    return interface
 
 
 def get_location_from_config_or_args(args, config) -> str:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@
 
   [project.optional-dependencies]
     test = ["pytest", "numpy>=1.5.0", "pytest-subprocess==1.5.0", "pytest-cov"]
+    types = ["mypy", "types-PyYAML", "types-redis", "types-requests", "types-ujson"]
 
   [project.urls]
     Home = "https://github.com/GreenScheduler/cats"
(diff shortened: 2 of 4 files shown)


tlestang:
Thanks for the work. The benefit of enforcing 100% compliance with mypy's static type checks is not clear to me. I think it's true of most projects big and small, but this becomes especially plain for a small simple projects like cats. Enforcing 100% will inevitably lead to code that aims at satisfying type checks at the expense of understandability. 

To me, this is demonstrated by the fact you felt the need to add a lengthy comment justifying why `configure.get_runtime_config` does not set `PUE`and `jobinfo` to `None`. I think `None` have different meaning than `0.0` and that's important. You also mention similar issues in your commit messages:

> This is arguably one of those places where static type checking and python do not play particularly well together, but this does allow mypy to report all is well.

> While None being passed into sys.exit isn't an error in python, it does break static type checking.

> While this isn't an error in python, it does break static type checking.

Don't get me wrong: i'm not against type hints, they can really help navigate the code. I'm not against static type checking either,
it can help identify issues not caught by tests, and it did in you case. But I do have a problem with (blindly?) enforcing 100% compliance with static type checks, for... what reason(s)?

andreww:
Thanks - I don't disagree about the balance between dynamic typing and including types, but would ask two questions. What is the point of annotating the types if we don't use that information? Is it sensible to include incorrect type annotations?

I would argue that _if_ we are adding typing information to functions it really ought to be correct. Running the code through mypy in the CI helps us make sure that the typing is correct, and adds some value to including the typing information (because we finds bugs that are not caught by our existing test suite). Otherwise we should probably strip out the type annotations.

Which isn't to say that the way I've made the static type checking pass is the best or only approach. Having said that, some context:

> While None being passed into sys.exit isn't an error in python, it does break static type checking.

We were using the returned value from a function without a `return` statement and feeding that into `sys.exit`. This is not an error in python but I'm not sure it's a sensible thing to do.

> While this isn't an error in python, it does break static type checking.

The variable name `output` was referring to to two different objects in the same place in the code. While this isn't an error in python, I think that it is potentially confusing.

pyproject.toml
```diff
@@ -18,6 +18,7 @@
 
   [project.optional-dependencies]
     test = ["pytest", "numpy>=1.5.0", "pytest-subprocess==1.5.0"]
+    types = ["mypy", "types-PyYAML", "types-redis", "types-requests", "types-ujson"]
```
abhidg, line 21: types-redis and types-ujson? are we using them?

abhidg:
We can have the best of both worlds - have mypy run but use `# type: ignore` annotations where we want be looser, or use typing.Any, or as @andreww suggested not use type annotations for a particular function. mypy can also ignore entire files.

pyproject.toml
andreww: Yes - indirectly and we need to pip install both of them them (we get errors if we don't). Both provide stubs for dependencies of requests-cache (which we use directly). My guess is that all four of these requirements stub packages will go away in time because they'll get merged with the main packages.

tlestang (approved):
Sorry @andreww , I feel like I shouldn't have quoted your commit messages out of context. I don't have anything against your changes (except maybe a minor readability comment below). Just wanted to question de use of `mypy` as-is in the CI, as I'm used to take its reports only as an indication that something might be wrong.. but maybe is fine too.

cats/configure.py
```diff
@@ -91,13 +97,14 @@ def CI_API_from_config_or_args(args, config) -> APIInterface:
...
     try:
-        return API_interfaces[api]
+        interface = API_interfaces[api]
     except KeyError:
         logging.error(
             f"Error: {api} is not a valid API choice. It must be one of " "\n".join(
                 API_interfaces.keys()
             )
         )
+    return interface
```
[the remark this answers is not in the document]
andreww: Yeh - the alternative is to return None (or not return anything) and then mypy wants us to check (e.g. with a "assert ... is not None") before we pass the returned value into something that needs interface to be the correct type. It is possible (but not sensible) to tell the logging object to not terminate when it sees an error, so I guess the test wouldn't be _totally_ redundant.

andreww:
I rebased this and found (what I think is) a better approach for the types of get_runtime_config. I'll add tests to this PR to make codeconv happy.

PR #85, opened 2024-04-17
Have tests report coverage in CI

andreww:
This will help us keep track of current test coverage - it's a simple change to the CI and dependancies. One thing to (maybe) think about is how these get reported (e.g. can we make github cache the previous coverage and use that to see if we are doing better). I'm happy for this to be merged or to leave it hanging around with more work needed.


abhidg (approved):
Thanks @andreww! we can use codecov to report coverage, I can set that up.

tlestang (approved):

PR #86, opened 2024-04-19
ci: upload coverage to codecov

abhidg:
This will need somebody with admin access to set the `CODECOV_TOKEN` in GitHub Secrets from https://app.codecov.io/github/GreenScheduler/cats/settings


ljcolling (approved):
I've added the token to the secrets so it LGTM

Issue #87, opened 2024-04-19
docs: remove `--jobinfo` and document config.yml


abhidg:
- [ ] Add sample `config.yml` to the main branch
- [ ] Remove `--jobinfo` from documentation

PR #90, opened 2024-04-22
Remove `--jobinfo`, standalone `config.yml` file

abhidg:
- docs: link to generated Sphinx documentation #87
- docs: remove --jobinfo and create standalone config.yml

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -3,8 +3,9 @@ import json
 import logging
 import subprocess
 import sys
-from argparse import ArgumentParser
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from datetime import timedelta
+from pathlib import Path
 from typing import Optional
 
 from .carbonFootprint import Estimates, get_footprint_reduction_estimate
@@ -17,6 +18,10 @@ from .optimise_starttime import get_avg_estimates  # noqa: F401
 # To add a scheduler, add a date format here
 # and create a scheduler_<new>(...) function
 SCHEDULER_DATE_FORMAT = {"at": "%Y%m%d%H%M"}
+
+
+def indent_lines(lines, spaces):
+    return "\n".join(" " * spaces + line for line in lines.split("\n"))
 
 
 def parse_arguments():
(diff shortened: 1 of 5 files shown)


cats/__init__.py
```diff
@@ -48,24 +54,41 @@ def parse_arguments():
...
+    carbon intensity of calculations on smaller computers. For example, the
+    above calculation can be scheduled by running:
 
         cats -d 90 --loc OX1 -s at -c 'mycommand'
+
+    To report carbon footprint, pass the `--config` option to select a
+    configuration file and the `--profile` option to select a profile. An
+    example config file is given below:
+
+{config_text}
```
tlestang, line 81: Very neat. Is that some sort of sphinx include syntax?

cats/__init__.py
```diff
@@ -48,24 +54,41 @@ def parse_arguments():
...
+    Examples
+    ********
 
-    Cats can be used to report information on the best time to run a calculation and the amount
-    of CO2. Information about a 90 minute calculation in centeral Oxford can be found by running:
+    CATS can be used to report information on the best time to run a calculation
+    and the amount of CO2. Information about a 90 minute calculation in centeral
+    Oxford can be found by running:
 
-        cats -d 90 --loc OX1 --jobinfo="cpus=2,gpus=0,memory=8,partition=CPU_partition"
```
tlestang, line 58: Oops, I forgot to remove this reference to the old 'jobinfo' syntax. Good catch :)

pyproject.toml
```diff
@@ -4,6 +4,7 @@
 
 [tool.setuptools]
   packages = ["cats"]
+  package-data.cats = ["config.yml"]
```
tlestang, line 7: I think `config.yml` would end up in `lib/python3.X/site-packages/cats` ? Might not be straightforward to use it from there?

tlestang (approved):

cats/__init__.py
abhidg: Just the standard f-string with the file read in

Issue #88, opened 2024-04-19
Remove unused carbon footprint files


abhidg:
Files like `fixed_parameters.yaml` are not being used anymore and should be removed.

PR #95, opened 2024-04-25
Remove unused carbon footprint files

abhidg:
Fixed parameters are not being used anymore as carbon footprint
calculations for flights etc. are not reported.

As these are fixed, these should be included in the carbon footprint
module when the functionality is re-added.

Fixes: #88


andreww (approved):
Looks straightforward

andreww:
I think the tests won't run because you've not changed any python files?

PR #96, opened 2024-05-01
Run tests on mac and python 3.12

andreww:
Our wrapping of "at" should run on macOS and with
python 3.12, so lets add these to the test matrix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,11 +19,12 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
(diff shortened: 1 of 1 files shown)


andreww:
Actually - something isn't right here, we have three extra tests "build" that I don't understand.

abhidg (approved):
Thanks @andreww - the checks that were 'running' were required checks set in branch protection rules. The required checks have to be present with the [exact name](https://github.com/orgs/community/discussions/26698#discussioncomment-4223158); since we did not have a 'build' step, but a 'build (3.9, ubuntu-latest)' step, GitHub was blocking the merge, since there was a required build step that was not running. I have changed the branch protection rules now.

andreww:
Thanks! - I learn something every day

PR #97, opened 2024-05-10
Change CLI argument name in Quickstart documentation

nicolaspayette:
At a glance, it looks like `--postcode` should be `--location`.


sadielbartholomew (approved):
Good spot, that is indeed the case. Thanks @nicolaspayette!

PR #98, opened 2024-05-17
Bump version to 1.0rc1, add authors

abhidg:


sadielbartholomew (approved):
All good, thanks Abhishek. (I ran this through a TOML and then pyproject.toml validator since I thought there was a small chance it wouldn't be happy with the accent character in Loïc's name, but it seems it is supported.)

andreww:
Just been down a rabbit hole about version numbers (is 1.0rc1 okay, or should we use 1.0.0rc1)? Turns out (according to https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) that they should be the same. So I'm happy

andreww (approved):

Issue #99, opened 2024-05-20
Errors when duration close to 48h


abhidg:
```
> cats -d 2820 --loc OX1
WARNING:root:config file not found
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk
Traceback (most recent call last):
  File "/Users/abhidg/.local/bin/cats", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/abhidg/.local/pipx/venvs/climate-aware-task-scheduler/lib/python3.12/site-packages/cats/__init__.py", line 284, in main
    now_avg, best_avg = get_avg_estimates(CI_forecast, duration=duration)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abhidg/.local/pipx/venvs/climate-aware-task-scheduler/lib/python3.12/site-packages/cats/optimise_starttime.py", line 26, in get_avg_estimates
    return wf[0], min(wf)
                  ^^^^^^^
ValueError: min() iterable argument is empty
```

```
> cats -d 2860 --loc OX1
WARNING:root:config file not found
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk
Traceback (most recent call last):
  File "/Users/abhidg/.local/bin/cats", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/abhidg/.local/pipx/venvs/climate-aware-task-scheduler/lib/python3.12/site-packages/cats/__init__.py", line 284, in main
    now_avg, best_avg = get_avg_estimates(CI_forecast, duration=duration)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abhidg/.local/pipx/venvs/climate-aware-task-scheduler/lib/python3.12/site-packages/cats/optimise_starttime.py", line 25, in get_avg_estimates
    wf = WindowedForecast(data, duration, start=datetime.now().astimezone())
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/abhidg/.local/pipx/venvs/climate-aware-task-scheduler/lib/python3.12/site-packages/cats/forecast.py", line 68, in __init__
[...]

PR #100, opened 2024-05-24
fix off by one

abhidg:
This fixes off-by-one errors in the code which led to #99. Detailed info is in commit messages.


andreww:
This all looks good to me but I think it would be good to turn the error in #99 into a test case - do we actually need a handful of integration tests where we feed in chosen forcasts and test the whole code?

ljcolling (approved):

PR #102, opened 2024-07-05
release v1

abhidg:
- Format config in example as a code block
- docs: fix typo
- Bump version to 1.0.0, add classifiers

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,7 @@ authors = (
 author = authors
 copyright = f"2023, {authors}"
 # The full version, including alpha/beta/rc tags
-release = "0.1.0"
+release = "1.0.0"
 
 # Add media: image for logo and a favicon for the browser tab
 html_logo = "_static/cats_dalle_img_200x200px_for_logo.png"
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -12,6 +12,8 @@
 import os
 import sys
 
+import cats
+
 sys.path.insert(0, os.path.abspath("../.."))  # cats repo root directory
 
 
@@ -26,7 +28,7 @@ authors = (
 author = authors
 copyright = f"2023, {authors}"
 # The full version, including alpha/beta/rc tags
-release = "1.0.0"
+release = cats.__version__
 
 # Add media: image for logo and a favicon for the browser tab
 html_logo = "_static/cats_dalle_img_200x200px_for_logo.png"
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -8,4 +8,10 @@ Install CATS via pip as follows:
 .. code-block:: console
    :caption: *Command to install CATS with pip.*
 
+   $ pip install climate-aware-task-scheduler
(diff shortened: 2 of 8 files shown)


docs/source/conf.py
```diff
@@ -26,7 +26,7 @@
 author = authors
 copyright = f"2023, {authors}"
 # The full version, including alpha/beta/rc tags
-release = "0.1.0"
+release = "1.0.0"
```
andreww, line 29: Not a problem for this PR, but I'll write it here so I don't forget as I review. Having to change the version string in two places feels like a recipe to forget one of them. Is it worth working out how to pull the version string out of the pyproject.toml file to use in the documentation configuration?

andreww (commented):
Changes look good to me

docs/source/conf.py
abhidg: Fixed in subsequent commits

pyproject.toml
```diff
@@ -25,12 +25,23 @@
   license = { file = "LICENSE" }
   requires-python = ">=3.9"
   readme = "README.md"
-  classifiers = ["License :: OSI Approved :: MIT License"]
+  classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python :: 3",
+    "Development Status :: 4 - Beta",
```
sadielbartholomew, line 31: If we're marking this as version 1.0.0, really we should bump this now to:

```suggestion
    "Development Status :: 5 - Production/Stable",
```

I think having the current state marked as both version 1.0.0 and 'beta' would give out mixed messages and therefore be confusing.

docs/source/installation.rst
```diff
@@ -8,4 +8,10 @@ Install CATS via pip as follows:
 .. code-block:: console
    :caption: *Command to install CATS with pip.*
 
+   $ pip install climate-aware-task-scheduler
```
sadielbartholomew, line 11: Is this meant to work right now? Because I get:

```console
pip install climate-aware-task-scheduler                                         ─╯
ERROR: Could not find a version that satisfies the requirement climate-aware-task-scheduler (from versions: none)
ERROR: No matching distribution found for climate-aware-task-scheduler
```

Or am I doing something wrong?

sadielbartholomew (commented):
Looks good overall but I have some comments.

docs/source/installation.rst
sadielbartholomew: (I should add, this wasn't using the branch here and assuming we can grab the library via PyPI directly.)

pyproject.toml
abhidg: Good point!

docs/source/installation.rst
abhidg: It is not yet uploaded to PyPi - it will work when I push the v1 release there

sadielbartholomew: OK that makes sense, thanks. Obviously we want to make sure it works ASAP after we officially commit that as the install mechanism, though.

PR #110, opened 2024-08-16
Adds option to specify config location using env variable

aturner-epcc:
Adds the option to set `CATS_CONFIG_FILE` environment variable to specify config file location. Config file location precedence becomes:

1. Command line option
2. Location specified in `CATS_CONFIG_FILE`
3. `./config.yaml`

### DIFF
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -12,6 +12,7 @@ runtime configuration consits of:
 
 import logging
 import sys
+import os
 from collections.abc import Mapping
 from typing import Any, Optional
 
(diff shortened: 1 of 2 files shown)


docs/source/quickstart.rst
andreww, line 1: I expect there will be a merge conflict for this file if #111 gets merged first. Should be easy to resolve.

cats/configure.py
```diff
@@ -76,7 +77,16 @@ def config_from_file(configpath="") -> Mapping[str, Any]:
...
-        # if no path provided, look for `config.yml` in current directory
+        # if no path provided, try to use config environment variable
+        cfile = os.getenv("CATS_CONFIG_FILE")
+        if cfile is not None:
+            try:
+                with open(cfile, "r") as f:
+                    return yaml.safe_load(f)
+                logging.info("Using config.yml found in CATS_CONFIG_FILE\n")
+            except FileNotFoundError:
+                logging.warning("CATS_CONFIG_FILE config file not found")
```
andreww, line 88: Would benefit from a test being added to test_configure.py (which will stop codecov throwing a fit). I'll see if I can add something to this PR for that.

andreww (approved):
Looks like a useful addition. It would be good to have a test case but I think that's easy given what we have.

cats/configure.py
andreww: I've created a PR against your main branch - I _think_ that will add a test to this PR if you merge it.

andreww:
I've rebased this and added a test. See PR #121 which contains this patch

PR #111, opened 2024-08-20
Update and clarify aspects in documentation

sadielbartholomew:
Minor updates to the CATS documentation, including:
* fix some typos appearing to the user. some misleading e.g. 'cpus' should be 'gpus' in the corresponding CLI argument description;
* clarify some statements which are somewhat misleading e.g. 'cats looks for a file named ``config.yaml`` in the current directory' when strictly it just needs a YAML extension so the same with `.yml` is also acceptable;
* update a code snippet implying that 'cats errors out' if no location is provided via CLI or configuration file, when actually at this stage CATS is more clever and will try to estimate location from the IP address.

Hopefully these are fairly self-explanatory suggestions, but please feel free to question.

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -66,7 +66,7 @@ def parse_arguments():
     ********
 
     CATS can be used to report information on the best time to run a calculation
-    and the amount of CO2. Information about a 90 minute calculation in centeral
+    and the amount of CO2. Information about a 90 minute calculation in central
     Oxford can be found by running:
 
         cats -d 90 --loc OX1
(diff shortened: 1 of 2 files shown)


andreww (approved):
Looks good to me. One thing I'm unclear about is if the documentation on readthedocs will keep up with our main branch, or keeps versions of the documentation for releases, or what. Not an issue for this PR, but probably needs looking at.

Anyhow, committing this quickly seems sensible.

andreww:
Go ahead and merge - that will tell us for sure what the merge conflicts are for #110. I think they would be easy to fix.

PR #112, opened 2024-08-23
Plot the carbon intensity time serise

andreww:
Just a proof of concept for now, but plots the CI forecast time series alongside our planned run times. 

Currently looks like this:

cats_plot

I think the optimal and now plots (green and red, respectively) are 30 mins out. Or background data is 30 mins out. Anyhow, this is a start.

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -14,6 +14,7 @@ from .CI_api_query import get_CI_forecast  # noqa: F401
 from .configure import get_runtime_config
 from .forecast import CarbonIntensityAverageEstimate
 from .optimise_starttime import get_avg_estimates  # noqa: F401
+from .plotting import plotplan
 
 __version__ = "1.0.0"
 
@@ -182,6 +183,11 @@ def parse_arguments():
         type=positive_integer,
         help="Amount of memory used by the job, in GB",
     )
+    parser.add_argument(
+        "--plot",
+        help="Create a plot of the forcast and optimised plan for the job",
+        action="store_true",
+    )
 
     return parser
 
@@ -312,6 +318,8 @@ This is usually due to forecast limitations."""
         print(output.to_json(dateformat, sort_keys=True, indent=2))
     else:
         print(output)
+    if args.plot:
+        plotplan(CI_forecast, output)
     if args.command and args.scheduler == "at":
         if err := schedule_at(output, args.command.split()):
             print(err)
--- a/cats/plotting.py
+++ b/cats/plotting.py
@@ -1,0 +1,38 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+def plotplan(CI_forecast, output):
+    """
+    Plot the carbon intensity forcast and optimised plan
+    """
+    # Just for now pull the CI forecast apart... probably belongs as method...
+    values = []
+    times = []
+    now_values = []
+    now_times = []
+    opt_values = []
+    opt_times = []
+    for point in CI_forecast:
+        values.append(point.value)
+        times.append(point.datetime)
+        if (point.datetime >= output.carbonIntensityOptimal.start and
+            point.datetime <= output.carbonIntensityOptimal.end):
+            opt_values.append(point.value)
+            opt_times.append(point.datetime)
+        if (point.datetime >= output.carbonIntensityNow.start and
+            point.datetime <= output.carbonIntensityNow.end):
+            now_values.append(point.value)
+            now_times.append(point.datetime)
+
+    # Make the plot (should probably take fig and ax as opt args...)
+    fig, ax = plt.subplots()
+    ax.fill_between(times, 0.0, values, alpha=0.2, color='b')
+    ax.fill_between(now_times, 0.0, now_values, alpha=0.6,  color='r')
+    ax.fill_between(opt_times, 0.0, opt_values, alpha=0.6, color='g')
+    ax.set_xlabel("Time (mm-dd hh)")
+    ax.set_ylabel("Forecast carbon intensity (gCO2eq/kWh)")
+    ax.grid(True)
+    ax.label_outer()
+    fig.autofmt_xdate()
+    plt.show()
+    return None


abhidg:
@andreww one way to handle matplotlib is by putting in as an optional dependency, in a `plots` group as an example. Then if users want the plotting functionality they can use `pip install 'climate-aware-task-scheduler[plots]'`

andreww:
> @andreww one way to handle matplotlib is by putting in as an optional dependency, in a `plots` group as an example. Then if users want the plotting functionality they can use `pip install 'climate-aware-task-scheduler[plots]'`

Thanks - I did this. It ended up being a little fiddly because of the type testing but it seems to be working. Ready to merge?

abhidg:
@andreww I don't see any interpolation here, running with `cats -d 60 --loc OX1 --plot`, I get this output:

```

The.____ ..... __ .... ________ . ______...
.. /  __)...../  \....(__    __).)  ____)....
..|  /......./    \......|  |...(  (___........
..| |limate./  ()  \ware.|  |ask.\___  \cheduler
..|  \__...|   __   |....|  |....____)  )....
...\    )..|  (..)  |....|  |...(      (..

WARNING:root:config file not found
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk

Best job start time                       = 2025-06-08 10:40:44
Carbon intensity if job started now       = 76.08 gCO2eq/kWh
Carbon intensity at optimal time          = 63.87 gCO2eq/kWh
```

Zooming in on the plots figure, I see a 30 min window highlighted, rather than 60 minutes (starting now, red):
Screenshot 2025-06-06 at 15 15 15

and the optimal time (green)

Screenshot 2025-06-06 at 15 16 14

andreww:
Thanks again for the spot @abhidg - fixed now.

sadielbartholomew:
@andreww @abhidg would a second review be useful at this stage, so we can get this merged - unless this is waiting on other work / something else?

sadielbartholomew:
OK sure @andreww I will get that done with review submitted this afternoon. Thanks for your work on this, it will be a really cool feature!

cats/__init__.py
```diff
@@ -193,6 +194,13 @@ def positive_integer(string):
         action="store_true",
         help="Disable all terminal output colouring (alias to --no-colour)",
     )
+    parser.add_argument(
+        "--plot",
+        help="Create a plot of the forecast and optimised plan for the job."
+        "This needs matplotlib to be installed, e.g. install with"
```
sadielbartholomew, line 200:
```suggestion
        help="Create a plot of the forecast and optimised plan for the job. "
        "This needs matplotlib to be installed, e.g. install with "
```

to avoid lack of spaces in the output e.g. "job.This"

cats/plotting.py
```diff
@@ -0,0 +1,67 @@
...
+
+    ax.text(0.125, 1.075, f"Mean carbon intensity if job started now: {output.carbonIntensityNow.value:.2f} gCO2eq/kWh",
+             transform=ax.transAxes, color='red')
+    ax.text(0.125, 1.025, f"Mean carbon intensity at optimal time: {output.carbonIntensityOptimal.value:.2f} gCO2eq/kWh",
+             transform=ax.transAxes, color='green')
+
+    ax.set_xlabel("Time (mm-dd hh)")
+    ax.xaxis.set_major_formatter(mdates.DateFormatter("%d-%m-%y %H:%M"))
+    ax.xaxis.set_minor_formatter(mdates.DateFormatter("%d-%m-%y %H:%M"))
+    ax.set_ylabel("Forecast carbon intensity (gCO2eq/kWh)")
```
sadielbartholomew, line 62: Since we have the power to do this in matplotlib as opposed to on the CLI where we can't format as such, it would be nice to apply formatting to the units e.g. for subscript using LaTeX for italicisation too. I suggest (this or similar):

```diff
diff --git a/cats/plotting.py b/cats/plotting.py
index 25feee3..53184b1 100644
--- a/cats/plotting.py
+++ b/cats/plotting.py
@@ -51,17 +51,18 @@ def plotplan(CI_forecast, output):
     ax.fill_between(now_times, 0.0, now_values, alpha=0.6,  color='r')
     ax.fill_between(opt_times, 0.0, opt_values, alpha=0.6, color='g')
 
-    ax.text(0.125, 1.075, f"Mean carbon intensity if job started now: {output.carbonIntensityNow.value:.2f} gCO2eq/kWh",
+    units_raw = r"$g\,CO_2eq/kWh$"
+    ax.text(0.125, 1.075, f"Mean carbon intensity if job started now: {output.carbonIntensityNow.value:.2f} {units_raw}",
              transform=ax.transAxes, color='red')
-    ax.text(0.125, 1.025, f"Mean carbon intensity at optimal time: {output.carbonIntensityOptimal.value:.2f} gCO2eq/kWh",
+    ax.text(0.125, 1.025, f"Mean carbon intensity at optimal time: {output.carbonIntensityOptimal.value:.2f} {units_raw}",
              transform=ax.transAxes, color='green')
 
     ax.set_xlabel("Time (mm-dd hh)")
     ax.xaxis.set_major_formatter(mdates.DateFormatter("%d-%m-%y %H:%M"))
     ax.xaxis.set_minor_formatter(mdates.DateFormatter("%d-%m-%y %H:%M"))
-    ax.set_ylabel("Forecast carbon intensity (gCO2eq/kWh)")
+    ax.set_ylabel(f"Forecast carbon intensity ({units_raw})")
```

which produces units like shown.
ev_units_latex

Though if preferred I can put in a separate PR to update some stylistic elements - I know you were asking about someone updating the plot that way (though it looks good enough already to me!).

cats/plotting.py
```diff
@@ -0,0 +1,67 @@
...
+    ax.fill_between(times, 0.0, values, alpha=0.2, color='b')
+    ax.fill_between(now_times, 0.0, now_values, alpha=0.6,  color='r')
+    ax.fill_between(opt_times, 0.0, opt_values, alpha=0.6, color='g')
+
+    ax.text(0.125, 1.075, f"Mean carbon intensity if job started now: {output.carbonIntensityNow.value:.2f} gCO2eq/kWh",
+             transform=ax.transAxes, color='red')
+    ax.text(0.125, 1.025, f"Mean carbon intensity at optimal time: {output.carbonIntensityOptimal.value:.2f} gCO2eq/kWh",
+             transform=ax.transAxes, color='green')
+
+    ax.set_xlabel("Time (mm-dd hh)")
```
sadielbartholomew, line 59: I am seeing instead datetimes corresponding to:

```suggestion
    ax.set_xlabel("Time (yy-mm-dd hh:mm)")
```

sadielbartholomew (approved):
Fantastic stuff, thanks @andreww! Works very well, as tested across various durations across the valid range and with a selection of curves from choosing random postcodes for location.

A few general questions before we merge (plus some in-line suggestions):

1. Why is there always a small gap before the curve is coloured red to indicate the 'immediate' start time? I would expect it to be red from the very leftmost point on the curve, but I might be missing something...

2. Overlapping of the immediate start curve region and the optimal one: I like how, if these overlap there is a different colour to indicate clearly enough that the red and green both cover that region (even if the khaki itself is a bit ugly!), like so:

evidence4

However, there are cases where the optimal time _is_ the immediate time (probably not too uncommon a case, since of about 10 runs I tried to test across duration and location inputs I saw this once), the whole thing is that colour which might be a bit confusing, i.e. if a user just sees this they can perhaps realise what is happening, but could easily not:

evidence2

Maybe a good way to solve that is to have the crossover region as a [hatched pattern](https://matplotlib.org/stable/gallery/shapes_and_collections/hatch_style_reference.html) over the other region e.g. green dots on the red? Alternatively, just use a key and include the colour of the overlap region.

Happy to bump this one (and e.g. the units formatting as mentioned in-line) to a new PR that I can quickly put in though, if you prefer, in the interests of getting this great work merged!

cats/plotting.py
andreww: I think this is best in a separate PR. In particular, I'm not 100% sure that pip install matplotlib gives you the LaTeX processor on all systems so it would be good to cook up a test case for that (and that isn't something that I'm proposing to do on this PR - CI with figure generation looks possible but a bit of a faff).

andreww:
Thanks @sadielbartholomew - I'll merge this and open an issue for improvements re. units markup and styling the overlap. In terms of:

> Why is there always a small gap before the curve is coloured red to indicate the 'immediate' start time? I would expect it to be red from the very leftmost point on the curve, but I might be missing something...

I think this is because I always plot the whole of the carbon intensity forecast which starts at the start of the "current" 30 min time window but the "now" time is always now. This means there could be up to 29 minutes of stale forecast in the plot. It wouldn't be too hard to trim this off but I'm not sure if it's more confusing to show the full forecast or have our initial point be different to what a naive reading of data on the front page of carbonintensity.org.uk says.

PR #113, opened 2024-08-26
Improve command output with text banner, colour and alignment

### LINKED ISSUE #105: Small ASCII 'CATS' text banner for start of STDOUT
Obviously not a critical-path priority piece of work, but I think it is really nice (striking, unique, dare I say fun), when terminal programs have a logo/banner which displays the library name at the start of the terminal output, e.g. as for or . I think we should add one for CATS, but this may be a controversial opinion, so let's treat this as a suggestion for now! (Though if there is agreement to go ahead, I can easily set this up.)

The exact ASCII art text can be chosen at a later stage, with [ASCII generator sites like this one ](https://patorjk.com/software/taag) creating the art for one to choose from, but the idea is something like this which will be seen when someone runs cats (though perhaps we can enable it to now be shown if someone specifies via come CLI or config file option):

```console
$ cats -d 90 --loc RG2                                                                      ─╯
 __     __   _____  __  
/ /`   / /\   | |  ( (` 
\_\_, /_/--\  |_|  _)_) 

WARNING:root:config file not found
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk
Best job start time 2024-07-21 11:27:36.231307+01:00
Carbon intensity if job started now       = 85.47 gCO2eq/kWh
Carbon intensity at optimal time          = 55.16 gCO2eq/kWh

Use --format=json to get this in machine readable format
```

or some other stylistic text choice e.g:

```console
  __   _  ___ __ 
 / _| / \|_ _/ _|
( (_ | o || |\_ \
 \__||_n_||_||__/
                 
```

or:

```console
             __         
.----.---.-.|  |_.-----.
|  __|  _  ||   _|__ --|
|____|___._||____|_____|
```

etc. (showing examples on the smaller side since we don't want anything to overhwhelm/dominate the output).

Colours can also be set for standard-supported terminal colours, but we don't have to go crazy with it (not that I would mind adding a little colour, as long as it doesn't in any significant way ramp up the carbon required to run, of course). :smile:

sadielbartholomew:
Close #105, including adding primitive colour to the terminal STDOUT, which can be disabled with a new CLI option `--no-colour` (shorthand `-n`) like with other standard commands, which make use of colour by default, allowing the choice to not add colour at all.

With this PR we go from a (human-readable, machine `--format=json` output unchanged with no colour) output without an ASCII text banner, to one with one before any logging or printing output, so from this:

```console
$ cats --duration 480 --location "RG2"
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk
Best job start time 2024-08-26 23:52:13.779577+01:00
Carbon intensity if job started now       = 150.64 gCO2eq/kWh
Carbon intensity at optimal time          = 150.64 gCO2eq/kWh

Use --format=json to get this in machine readable format
```

to this (add banner, align figures, subsume the format info message into the logging system):

```console
$ cats --duration 480 --location "RG2"

The.____ ..... __ .... ________ . ______...
.. /  __)...../  \....(__    __).)  ____)....
..|  /......./    \......|  |...(  (___........
..| |limate./  ()  \ware.|  |ask.\___  \cheduler
..|  \__...|   __   |....|  |....____)  )....
...\    )..|  (..)  |....|  |...(      (..

WARNING:root:config file not found
WARNING:root:Unspecified carbon intensity forecast service, using carbonintensity.org.uk

Best job start time                       = 2024-08-27 10:14:59.692606+01:00
Carbon intensity if job started now       = 191.95 gCO2eq/kWh
Carbon intensity at optimal time          = 109.52 gCO2eq/kWh
```

though note I can't demonstrate the colours introduced through the a markdown code snippet, so here is a demonstrative screenshot of what output you get from the same command on the current `main` and on this branch:

Screenshot_20240827_000324_branch_compare

### Reviewing

I am happy to tweak any subjective elements e.g. the styling of the ASCII banner or the colours chosen, etc., but thought for progress it best put this PR up as a suggestion on these.

As for the banner design chosen for this starting point:

* I used the site linked in my Issue for the main lettering;
* took the 'Goofy' font option as my favourite choice;
* replaced whitespace in the background with dots which resemble plot gridlines to me and they make it somewhat customised and less generic (i..e not pulled as-is from the site I used);
* added to the dots at the end to form an arrow shape, to further customise it;
* added some colours on the lettering to make it stand out more.

I could go a little crazier with the colours but thought it best to keep it fairly minimal mostly to highlight any green as 'good' (or strictly, 'better') and red as 'bad' ('worse') for the figures as well as a little, but less so, emphasis on the CATS banner text.


sadielbartholomew:
Looks like I need to update the tests - so moving this PR back to draft form. I'll do that this evening hopefully then re-open for review.

abhidg:
Looks great! I tested on a light terminal and it works okay there as well. Windows might not work - but I think the main use case of CATS is on Linux machines anyway.

image

sadielbartholomew:
Thanks @abhidg for the review. Very good point, I didn't think to check how the output looks on a light-on-dark (light background) colour scheme, since I always use light-on-dark. What you show in your screenshot looks fine in my view, too, though I think I should probably change the colour of the 'CATS' capital lettering to be the default colour rather than white (so black in your screenshot case, like the plain text), to ensure it is always visible.

In the next few days I shall get the tests updated here and add some testing of the new `--no-colour` option, then request a final review, but anyone who wants to (re-)review before or after the testing updates is welcome to.

cats/__init__.py
```diff
@@ -182,6 +183,12 @@ def positive_integer(string):
         type=positive_integer,
         help="Amount of memory used by the job, in GB",
     )
+    parser.add_argument(
+        "-n",
+        "--no-colour",  # or could use American spelling like many other CLIs?
```
andreww, line 188: can we accept colour and color? I think add_argument accepts any number of names so just adding "--no-color" ought to work.

andreww (approved):
Yep - looks good to me - I like it. Couple of questions. 

First, what does this do for people who use screen readers in a terminal? I've no idea if people do use screen readers for working in a terminal (or, indeed, who to ask for advice on this) but I do wonder if an option to revert back to not using a text banner could help with accessibility. Actually, there seems to be one study (it claims to be the only one) at https://dl.acm.org/doi/fullHtml/10.1145/3411764.3445544. Section 5.4 may be relevant. Having said that, I suspect our json output format is already a big win for that. I'm agnostic about this.

Second, do you think it's worth putting the option to turn the banner and/or the colours off in the configuration file (or should we just expect that people know how shell aliases work)? Again, I'm agnostic.

sadielbartholomew:
Thanks @andreww for the review. Very good point RE accessibility.

> I do wonder if an option to revert back to not using a text banner could help with accessibility.

Given this thought I am wondering if we could perhaps replace the 'no colour' option I've used in this PR as-is with a more general 'minimal' (`--minimal-view` or similar) or 'plain' (`--plain` or similar) option which would both remove the colouring and the text banner? And, as a bonus, prevent us from having to wrestle with the American/British spelling of colour issue?

> Second, do you think it's worth putting the option to turn the banner and/or the colours off in the configuration file (or should we just expect that people know how shell aliases work)? Again, I'm agnostic.

Good point also. I will look into the means to specify via the configuration file when I update this PR, hopefully tomorrow. Though since it concerns the command output itself, I personally think it is OK to have it just as something for the CLI, unless all of the other options are possible to set both from a config. file or the command-line, in which case I don't want to create some weird special case.

I'll report back once I update the PR.

cats/__init__.py
sadielbartholomew: Good idea, I've implemented this now, so marking as resolved.

sadielbartholomew:
Bump @GreenScheduler/cats team - this PR has been sitting there a while now but is ready to go in my eyes - it has already had a few reviews and I have responded to all feedback after those so just needs a quick re-review to ensure the new commit changes are OK and/or an overall sanity check.

abhidg:
Thanks, looks good!

PR #121, opened 2024-12-06
PR 110 redux

andreww:
@aturner's PR #110 with merge conflict resolved and a test added

As previously:  Adds the option to set `CATS_CONFIG_FILE` environment variable to specify config file location. Config file location precedence becomes:

1. Command line option
2. Location specified in `CATS_CONFIG_FILE`
3. `./config.yaml`

### DIFF
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -12,6 +12,7 @@ runtime configuration consits of:
 
 import logging
 import sys
+import os
 from collections.abc import Mapping
 from typing import Any, Optional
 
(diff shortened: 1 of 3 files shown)


andreww:
All looked okay in PR #110 - could somebody review and merge?

abhidg (commented):
Looks good! Doesn't look like CATS actually loads `config.yaml` (only `config.yml`) though, so we should change the docs to match

abhidg (approved):

PR #125, opened 2025-03-04
122 sbatch

### LINKED ISSUE #122: scheduler: add sbatch

abhidg:
- **ci: update pre-commit hooks**
- **feat: add sbatch scheduler support, fixes #122**

### DIFF
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -263,9 +263,7 @@ def print_banner(disable_colour):
         print(CATS_ASCII_BANNER_COLOUR)
 
 
-def schedule_at(
-    output: CATSOutput, args: list[str], at_command: str = "at"
-) -> Optional[str]:
+def schedule_at(output: CATSOutput, args: list[str]) -> Optional[str]:
     """Schedule job with optimal start time using at(1)
 
     :return: Error as a string, or None if successful
@@ -274,7 +272,7 @@ def schedule_at(
     try:
         subprocess.check_output(
             (
-                at_command,
+                "at",
                 "-t",
                 output.carbonIntensityOptimal.start.strftime(
                     SCHEDULER_DATE_FORMAT["at"]
--- a/cats/__init__.py
+++ b/cats/__init__.py
@@ -20,7 +20,7 @@ __version__ = "1.0.0"
 
 # To add a scheduler, add a date format here
 # and create a scheduler_<new>(...) function
-SCHEDULER_DATE_FORMAT = {"at": "%Y%m%d%H%M"}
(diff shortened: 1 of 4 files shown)


cats/__init__.py
```diff
@@ -290,6 +289,29 @@ def schedule_at(
         return f"Scheduling with at failed with code {e.returncode}, see output below:\n{e.output}"
 
 
+def schedule_sbatch(output: CATSOutput, args: list[str]) -> Optional[str]:
```
andreww, line 292: There is an interesting difference between this signature and the schedule_at signature - for at we allow (in the code, but the user cannot access) alternative choices to the default "at" command to be passed in. Here you don't, "sbatch" is hard coded below. I think this is fine, but it's probably worth a comment. (TBH, I'm not clear why we go through the Popen and piping process for at either - but do remember that this was a PITA to get working).

andreww (approved):
This all looks okay to me. Has it been tested in reality on a system running slurm?

cats/__init__.py
abhidg: Not sure why we allow at_command to be overridden, AFAIK all POSIX systems should have the same name for at. I will remove the at_command from the schedule_at() call. As for the piping for at, that is because at accepts the command name through stdin.

andreww: I think it gets used in the tests - to see what happens when at is broken - but there are other ways of doing that I guess.

sadielbartholomew:
@abhidg do you need a second review or is one sufficient?

abhidg:
@sadielbartholomew could you check if this PR works in a cluster? I don't have an account on our supercomputer ATM.

sadielbartholomew:
> could you check if this PR works in a cluster? I don't have an account on our supercomputer ATM.

Sure, as long as you don't mind waiting a couple of days - I have some urgent work to sort for tomorrow and it will take a bit more time to review on JASMIN or ARCHER2. Will aim to submit a cluster-based review by the end of the week.

colinsauze:
I've just tested this on our cluster at NOC and it is happily submitting jobs to the slurm queue.
To match the behaviour of invoking sbatch directly (and how Cats works with At) it would be very helpful to be able to display the job number on screen after submission.

colinsauze:
Also tested on JASMIN (newer version of slurm, 24.11.0 instead of 23.02.6) and it is working there.

abhidg:
Thanks @colinsauze! I have created an issue to display the job number on submission.

sadielbartholomew:
Sorry @abhidg this clearly fell off my radar, but thanks @colinsauze for testing on JASMIN! Brilliant to hear that it is working with SLURM.

PR #129, opened 2025-03-28
ci: disable codecov patch and project status checks

abhidg:


andreww (approved):
Looks sensible - just turning this off

PR #132, opened 2025-03-31
Update author details

andreww:
Add Thibault's details


andreww:
Two people didn't reply - and the list is alphabetical - but I'll merge this for now.

PR #138, opened 2025-04-15
SLB minor feedback on paper draft

sadielbartholomew:
Here is my draft feedback which are only minor/copy-edit points. These suggestions, mostly:

* fix some typos and minor grammatical issues
* avoid the use of words such as 'simple' or 'straightforward' which are subjective and not very specific, by replacing with words to the effect that are closer to the context of the statements in question (please amend as you wish, but I think we should avoid use of 'simple' etc. by some means)
* attempt to reduce the amount of parenthesised statements, especially in paragraphs one (summary section paragraph) and three (second in the statement of need section) - since there are lot of these to provide extra context e.g. through clarifying terms/acronyms or referencing concrete case studies etc., I think it would be best to use other punctuation or tweak to avoid use of them for other clauses

Please take the suggestions you agree with and feel free to leave anything you don't, and if you want clarification as to why I made a particular suggestion, just ask against relevant line(s).

(Overall I do think we are likely to be told it is too long and to shorten the text, but I think we can cross that bridge if we get to it.)


andreww (approved):
Yes - all looks good to me - thanks!

PR #139, opened 2025-05-02
Add material to documentation

andreww:
We've been asked to shorten the JOSS manuscript. I think the best bet is to remove the "how this works" material from the manuscript and put it into the documentation. This PR captures that work.

Thus far, this is just text lifted from the manuscript and converted into restructured text (if I have that correct).


andreww:
Does anybody know if / how we can see unmerged documentation changes? We should probably document that in the maintainers section!

abhidg:
There isn't an easy way to look at the PR without building locally. We could move to ReadTheDocs which provides a way to preview documentation in PRs: https://docs.readthedocs.com/platform/latest/pull-requests.html

andreww:
Yes - that would be good. Worth wrapping it all up in this PR?

andreww:
I've got Read The Docs working on PRs now, so I'll close this and reopen a new PR to see the changes.

Issue #137, opened 2025-04-14
[sbatch] display job number on screen after submission


abhidg:
I've just tested this on our cluster at NOC and it is happily submitting jobs to the slurm queue.
To match the behaviour of invoking sbatch directly (and how Cats works with At) it would be very helpful to be able to display the job number on screen after submission.

_Originally posted by @colinsauze in https://github.com/GreenScheduler/cats/issues/125#issuecomment-2802458070_

andreww:
Fixed by #137

PR #140, opened 2025-05-12
Display sbatch output

colinsauze:
Fixes issue #137 and displays the output of sbatch (e.g. "Submitted batch job 127"). 

It turns out that `at` displays the job number/status via stderr and that is what was causing it to appear on screen. `Sbatch` uses stdout, so this code captures and displays that output.


abhidg (approved):
Looks good!

PR #146, opened 2025-06-06
Allow cats_config.yml and cats_config.yaml

### LINKED ISSUE #145: Potential filename clashes with `config.yml`
When running without a `--loc` and without supplying a value to `--config`, CATS will look for `config.yml` in the current directory. As this is a generic file name, there's a chance that this could clash with a config file used for an unrelated purpose. Perhaps the default file name could be something like `cats-config.yml`? 

I'd also recommend checking for the `.yaml` file extension, as in testing I'd accidentally named mine `config.yaml` and it took me a while to figure out why it wasn't working. 

(this isn't part of my JOSS review, I just think CATS is pretty neat and wanted to help out!)

andreww:
This change allows us to read configuration data from local files cats_config.yml, cats_config.yaml or config.yml. The first file found (in that order) is read and any others are ignored.

I've also updated the tests to check that this works which involved fun with fixtures. Having thought about it a bit more this could probably be cleaned up with a parameterisation but I'm not sure about extracting the parameter for the test in the fixture ... something for another day.

Still need to update the documentation to match and as well as cleanup in the tests, I'm pretty sure that config_from_file function could be simplified. 

Fixes #145

### DIFF
--- a/cats/configure.py
+++ b/cats/configure.py
@@ -103,7 +103,7 @@ def config_from_file(configpath="") -> Mapping[str, Any]:
                 logging.warning("config file not found")
                 conf_dict = {}
 
-        return conf_dict
+    return conf_dict
 
 
 def CI_API_from_config_or_args(args, config) -> APIInterface:


cats/configure.py
```diff
@@ -83,14 +83,16 @@ def config_from_file(configpath="") -> Mapping[str, Any]:
             try:
                 with open(cfile, "r") as f:
                     return yaml.safe_load(f)
-                logging.info("Using config.yml found in CATS_CONFIG_FILE\n")
+                logging.info(f"Using {cfile} found in CATS_CONFIG_FILE\n")
             except FileNotFoundError:
                 logging.warning("CATS_CONFIG_FILE config file not found")
         # if no path provided and no env variable, look for `config.yml` in current directory
+        config_file_names = ["cats_config.yml", "cats_config.yaml", "config.yml"]
```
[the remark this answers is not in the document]
andreww: That's all done

andreww:
@sadielbartholomew, @abhidg  - thanks for the comments. I think this is now ready to review and merge.

cats/configure.py
```diff
@@ -76,26 +76,34 @@ def config_from_file(configpath="") -> Mapping[str, Any]:
...
-                return yaml.safe_load(f)
-            logging.info("Using config.yml found in current directory\n")
-        except FileNotFoundError:
-            logging.warning("config file not found")
-            return {}
+        else:
+            # if no path provided and no env variable, look for a file in current directory
+            # we support several file names but warn for deprecated names
+            config_file_names = ["cats_config.yml", "cats_config.yaml", "config.yaml"]
+            cfile = next((x for x in config_file_names if os.path.isfile(x)), "config.yml")
```
sadielbartholomew, line 95: Why don't we check that the 'config.yml' case is a file, like with the other three? This seems simpler and more comprehensible:

```suggestion
            config_file_names = ["cats_config.yml", "cats_config.yaml", "config.yaml", "config.yml"]
            cfile = next(x for x in config_file_names if os.path.isfile(x))
```

andreww: The reason I did it like this that we need something in that cfile variable (even if the file doesn't exist) so we can report that it doesn't exist. Otherwise we would need an extra test for None.

abhidg:
@andreww looks good to me!

cats/configure.py
sadielbartholomew: Ah OK that makes sense - you could potentially add in a comment to say as much so others don't wonder the same thing, but it is not a blocker. I'll resolve this conversation. I am happy and so is Abhishek so please merge when ready.

PR #147, opened 2025-06-08
ci: publish to PyPi on new release, fixes #141

### LINKED ISSUE #141: [ci] Setup publishing to PyPi on new tag

abhidg:


sadielbartholomew (approved):
By eyeball, looks to configure the dedicated publish external workflow correctly for our context and the YAML is all valid so nothing strikes me as amiss, however I don't know a good way to test it explicitly without changing it around too much to us TestPyPI - don't want to accidentally create a new official release!

Happy to approve regardless but for validation/reassurance it would be good to know it has been tested - have you checked this via publishing to TestPyPI or otherwise?

andreww (approved):
I think essentially we just need to try in and see. Worst that can happen? An extra version on PyPi that doesn't work and we fix it by hand. The process should probably be documented (but maybe just open an issue with text saying how to create a PyPi release.

I'm guessing there are also settings on the PyPi side that you've had to change. Are these tied to your account there @abhidg? We probably need to avoid the situation where we have single person points of failure (I guess I need to deal with the same thing on ReadTheDocs now).

Issue #157, opened 2025-07-08
MacOS changes in github actions


andreww:
I got an email from github pointing out that `macos-latest` is going to change from macOS 14 to macOS 15 in stages through August. Shouldn't bother us, but if it does we can run on macOS 14 with `runs-on: macos-14` (which will be maintained for at least a year) while we work out what's going wrong

PR #160, opened 2025-08-11
Add testing for python 3.13

andreww:
This is the current python stable relase. 3.9 goes out of security support in a month or two, so we should update the requierments then. This also gives us a chance to check the new MacOS test runners on github work for us.


andreww:
All github OS X runners were on version 14.7.6 just now, so #157 still needs to be checked

Issue #64, opened 2024-01-28
Ability to constrain the suggested start time to a specific time window within which the job should start


abhidg:
Suggested in #29

PR #166, opened 2025-10-05
feat: Implement window constraints with new command-line options.

### LINKED ISSUE #134: Support calculation over specified time < 48 hours by cutoff
At present, it seems CATS only allows users to determine the greenest time to run within the next **full 48 hour time period**, but I can easily foresee that they may instead want to determine the greenest time **within a set time period that is less than 48 hours** - say, to find out when the estimated carbon intensity is lowest over the next 24 hours only, or even for a shorter job over the next 3 hours only. We can support this, as long as the time window isn't so small it approaches the consistent (or maximum) interval time at which the forecasts are sampled (see #57), though it depends also on the expected job duration as to what is feasible.

An example would be a researcher who wants to set, just before they leave work in the evening, a job to run overnight before they start work again the next morning, and they don't care when otherwise so use CATS to schedule it for the greenest time - at the moment CATS might schedule it dependent on the forecast for the next day later than the morning or even the day after that, which wouldn't be useful for the researcher who needs it to be run for the next morning.

So, I am suggesting a way to input a _cutoff_ time or _maximum window_ time (those are the best ways I can think of right now to describe it, but happy to take suggestions) to reduce the maximum time of scheduling to a specified time T. This would mean performing the calculation to find the minimum carbon intensity over not 48 hours but the specified T. This should be simple enough to implement as an input parameter to reduce the time window to consider.

I suggest a name of either `--window` or `--cutoff` or similar for the new configuration input which would be a time value between some sensible minimum and 48 hours, in minutes, to provide the window in which to consider to search for the greenest start time within, which would by default be 2880 minutes (= 48 hours).

isofinly:
This PR closes #134 and #64

Features: 

1. Maximum Window Time (--window): Allows users to limit the forecast horizon from the default 48 hours to a specified duration T. Window size out of range (1-2880 minutes)
2. Time Window Constraints (--start-window, --end-window): Enables constraining the suggested start time to specific time windows

Supports different time zones, formats. Enforces start before end. 

Examples:

Basic Window Limitation
```bash
# Search only next 12 hours instead of 48
cats -d 60 --loc OX1 --window 720
```

Business Hours Constraint
```bash
# Job must start between 9 AM and 5 PM
cats -d 180 --loc SW7 --start-window "09:00" --end-window "17:00"
```

Maintenance Window
```bash
# Job must start during 2-6 AM maintenance window
cats -d 240 --loc M1 --start-window "2024-01-15T02:00" --end-window "2024-01-15T06:00"
```

Combined Constraints
```bash
# 12-hour search window + business hours constraint
cats -d 120 --loc OX1 --window 720 --start-window "09:00" --end-window "17:00"
```

With Schedulers
```bash
# Integration with AT scheduler
cats -d 90 --loc OX1 --start-window "09:00" --end-window "17:00" -s at -c "job.sh"

# Integration with SLURM
cats -d 240 --loc SW7 --window 720 -s sbatch -c "slurm-job.sh"
```

### DIFF
--- a/cats/CI_api_interface.py
+++ b/cats/CI_api_interface.py
@@ -47,7 +47,7 @@ def ciuk_parse_response_data(response: dict):
     and is set up to cache data from call to call even accross different
     processes within the same half hour window. The returned prediction data
     is in half hour blocks starting from the half hour containing the current
-    time extending up to 48 hours into the future.
+    time and extending up to 48 hours into the future.
 
     :param response:
     :return:


cats/__init__.py
```diff
@@ -201,6 +276,27 @@ def positive_integer(string):
...
+    parser.add_argument(
+        "--start-window",
+        type=str,
+        help="Earliest time the job is allowed to start, in ISO format (e.g., '2024-01-15T09:00'). "
+        "If only time is provided (e.g., '09:00'), today's date is assumed. "
+        "Timezone info is optional and defaults to system timezone.",
+    )
+    parser.add_argument(
+        "--end-window",
+        type=str,
```
andreww: Could change to `type=parse_time_constraint` - see above

cats/__init__.py
```diff
@@ -28,6 +31,78 @@ def indent_lines(lines, spaces):
...
+    :raises ValueError: If constraints are invalid
+    """
+    # Validate window minutes
+    if window_minutes < 1 or window_minutes > 2820:
+        raise ValueError("Window must be between 1 and 2820 minutes (47 hours)")
+
+    start_datetime = None
+    end_datetime = None
+
+    # Parse time constraints if provided
```
andreww: If you parse these in argparse, you won't need this logic here

tests/test_window_constraints.py
```diff
@@ -0,0 +1,534 @@
...
+        assert result.replace(tzinfo=None) == expected_time.replace(tzinfo=None)
+
+    def test_parse_empty_string_returns_none(self):
+        """Test that empty string returns None."""
+        result = parse_time_constraint("")
+        assert result is None
+
+    def test_parse_none_returns_none(self):
+        """Test that None returns None."""
+        result = parse_time_constraint("")  # Pass empty string instead of None
```
andreww: Why pass the empty string here - it turns out to be the same as the previous test?

PR #177, opened 2025-12-03
Include plot in introductory information to illustrate concept

sadielbartholomew:
After #171 updated the plot from POC form, it is nice to showcase it in the introductory information, with a brief note about how the optimal window is where the area under the curve is minimised, as a neat visual demonstration of the overall concept of carbon intensity minimisation which is the key concept for CATS. Namely:

* include in README;
* include in the 'quickstart' page of the docs (here the 'intro' is more about a summary and background, rather than any nitty-gritty, so best in 'quickstart' instead IMO).

Also moves the GIF to the `docs.source/_static` directory, updating any links to that path (just two), since all other images and media are there, as with the example plot which I generated and committed. Alternatively we could have a root-level 'media' directory, but with minimal media I don't think this is necessary.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -69,7 +69,10 @@ compared to at the optimal time. For example:
    Carbon intensity if job started now       = 217.41 gCO2eq/kWh
    Carbon intensity at optimal time          = 118.65 gCO2eq/kWh
 
-![CATS command run plot example output for RG1 and 3 hour job](docs/source/_static/example_plot_output_rg1_180mins.png)
+.. image:: _static/example_plot_output_rg1_180mins.png
+  :width: 400
+  :alt: CATS command run plot example output for RG1 and 3 hour job. The graph shows the forcast carbon intensity, a red region representing running the job now and a green region of lower intensity representing running the job in the future.
+  :align: center
 
 The optimal window is where the area under the curve is minimised, as
 highlighted in the plot ('Optimal job window').


docs/source/quickstart.rst
```diff
@@ -29,6 +29,22 @@ should be transparent to cluster users.
 It will display the time to start the job on standard out and optionally
 some information about the carbon intensity on standard error.
 
+Illustration of estimate with ``--plot``
```
andreww: I don't think it has to be in this commit, but since we've got one it would probably be worth showing the output plot somewhere. I wonder if it's also worth including the STDOUT output in the console block here and in the example above?

andreww:
Strikes me that we don't have "rules" written down anywhere about who should merge PRs, just a requirement to have them reviewed. I'm happy for @sadielbartholomew to hit merge for this but that's something we should probably document in contributing...

andreww (commented):
Everything looks good apart from the plot image in the rendered quickstart document (https://cats--177.org.readthedocs.build/en/177/quickstart.html) is showing (for me) as text and doesn't show the actual plot. I think this is a markdown v's restructured text issue. I'll take a look and see if I can quickly figure out the fix.

docs/source/quickstart.rst
andreww: Oh - and I think cropping the warnings is fine. If you had a local config file they would go away.

PR #180, opened 2026-03-10
Add and document a CATS User-agent

andreww:
This adds a User-agent header to all CATS HTTP API calls of:

    CATS/version +https://cats.readthedocs.io/

where the version is set according to __version__ in __init__. These headers are added to our calls to the carbon intensity and ipapi APIs.

Add documentation outlining our User-agent header, and our approach to caching.  This hope is that this will mean our calls to ipapi don't get bundled in with
python requests calls when any rate limiting is going on. This should also allow the carbon intensity team to see how much we are (or are not) using their service.

### DIFF
--- a/cats/CI_api_query.py
+++ b/cats/CI_api_query.py
@@ -3,7 +3,7 @@ from datetime import datetime, timezone
 import requests_cache
 
 from .forecast import CarbonIntensityPointEstimate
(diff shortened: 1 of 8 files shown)


abhidg (approved):
Looks good to me!

PR #185, opened 2026-03-26
Add windows support

andreww:
Currently we just add windows-latest to CI to see if we can run the test cases with the windows-latest CI test runner at github (which is windows server 2025 running on a x64 VM).

Just needed to add a conditional dependency on tzdata in the pyproject.toml file to make the tests go from failing to passing. It would be good to have a real windows user test it (but that'll need a pip install from this branch rather than from PyPi).

Also document python version and supported OSes.


abhidg (approved):
Looks good, thanks @andreww

PR #195, opened 2026-05-18
docs: remove deprecated conda env mention, fixes #194

### LINKED ISSUE #194: Update deprecated sections of docs/README.md
It doesn't seem like we use most of this now and readthedocs handles build on deployment. Specifically are we using conda to build locally? If not we can remove that section.

abhidg:


andreww (approved):
Looks good to me. I've not tried to carefully check that everything in that requirements file matches what we have in the [docs] dependencies in pyproject.toml. However, I have used pyproject.toml to build our docs and that's where I added the new dependancy for bibtex. Happy for you to merge this (and noting that this is also the kind of thing that we ought to include in a future "developers" section of the main documentation).

PR #196, opened 2026-06-28
Support other APIs and metric types using a new providers interface

### LINKED ISSUE #172: Headless mode access to plot by means to save
Out plotting functionality creates an illustrative plot for a given configuration, but as an end result it only opens this in a GUI window via (matplotlib's) `plt.show()` - which doesn't work in a headless context - notably when on remote servers like JASMIN using SSH or in automated scripts. So in these cases, there's no (non-contrived, e.g. without setting up some sort of GUI window support e.g. via X) way to access the plot. Which would become a major deficiency given CATS v2 is targeting HPC/HTC application.

To get around this problem, I suggest we implement the following:
* logic to detect if headless i.e. querying `os.environ.get("DISPLAY")` or similar, in which case save the plot to the current working directory with a default name;
* an extra config. option `--save-plot` taking a (path and) filename argument for saving the plot instead of opening it (upon which you do generally get given an option to save from the GUI widgets), as an optional override the user can set.

Happy to take other ideas, though!

abhidg:
This PR adds a providers interface that replaces the current `CI_api_query` and `CI_api_interface` classes. The advantage of the new approach is that it allows further customisation of the provider rather than customising the URL; this is useful to support providers that might need us to interface directly with a databases or with HTTP APIs requiring authentication. New types (`Timeseries`) have been added to support not just carbon intensity, but any other type of forecast timeseries with customizable units.

Other changes:
- Added type hinting throughout
- Using Exceptions for error handling (`cats.exceptions`) which allows us to consolidate error handling in one place (`main()`) as well as providing type based checking
- Added `--save-plot` to save plots to file

Remaining refactor would be to move configuration parsing and error handling out of cats/cli.py, this will be done in a subsequent PR.

### DIFF
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -1,29 +1,51 @@
 from datetime import datetime, timedelta
 
 from cats.forecast import (
-    CarbonIntensityAverageEstimate,
-    CarbonIntensityPointEstimate,
+    AverageEstimate,
+    PointEstimate,
     WindowedForecast,
 )
+
 
 def test_127_bug_report():
     """
     Regression test for the bug report in issue 127
-    As far as I can tell, the best option for the provided data is to 
+    As far as I can tell, the best option for the provided data is to
     start at once on a decreasing trend. However, our results do
-    differ to those reported in the issue. I think because of the 
+    differ to those reported in the issue. I think because of the
     fixes that went in before version 1.1.
     """
     # Reported input data
     d = datetime(year=2025, month=3, day=10, hour=15)
     data = [
-        CarbonIntensityPointEstimate(
-        datetime=d + timedelta(minutes=idt*30),
-        value=v,
-    )
-    for idt, v in zip(range(19),
-                      [14, 20, 20, 22, 13, 40, 49, 76, 91, 86, 
-                       61, 28, 24, 26, 25, 27, 27, 26, 23])
+        PointEstimate(
+            datetime=d + timedelta(minutes=idt * 30),
+            value=v,
+        )
+        for idt, v in zip(
+            range(19),
+            [
+                14,
+                20,
+                20,
+                22,
+                13,
+                40,
+                49,
+                76,
+                91,
+                86,
+                61,
+                28,
+                24,
+                26,
+                25,
+                27,
+                27,
+                26,
+                23,
+            ],
+        )
     ]
 
     # Reported job data
@@ -34,14 +56,14 @@ def test_127_bug_report():
 
     now_avg, best_avg = result[0], min(result)
 
-    interp1 = 20 + (2.0/30)*24 # interpolate onto earlest start time (now)
-    interp2 = 13 + (9.0/30)*6 # interpolate onto end time
-    expected_best = CarbonIntensityAverageEstimate(
+    interp1 = 20 + (2.0 / 30) * 24  # interpolate onto earlest start time (now)
+    interp2 = 13 + (9.0 / 30) * 6  # interpolate onto end time
+    expected_best = AverageEstimate(
         start=datetime(2025, 3, 10, 16, 24),
         end=datetime(2025, 3, 10, 16, 54),
-        value=(0.5*(interp1 + 22)*6 + 0.5*(22 + interp2)*24)/30 , # integral
-        start_value= interp1, 
-        end_value= interp2 
+        value=(0.5 * (interp1 + 22) * 6 + 0.5 * (22 + interp2) * 24) / 30,  # integral
+        start_value=interp1,
+        end_value=interp2,
     )
 
     assert now_avg == expected_best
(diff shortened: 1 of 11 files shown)


cats/providers/base.py
```diff
@@ -0,0 +1,79 @@
...
+    session = requests_cache.CachedSession("cats_cache", use_temp=True)
+    headers = headers or {}
+    headers.update(user_agent)
+    return session.get(url, headers=headers).json()  # pyright: ignore[reportUnknownMemberType]
+
+
+class BaseProvider(ABC):
+    "Base provider class from which API providers in CATS derive from"
+
+    MAX_DURATION_MINUTES: ClassVar[int]
```
andreww: Do we also need a temporal resolution here? I think carbonintensity gives data every 30 mins but wattnet uses 15 min resolution.

cats/providers/base.py
```diff
@@ -0,0 +1,79 @@
+"""
+Base API provider
+-----------------
+
+This module provides the base provider class from which all API providers for
+CATS are derived from. To define a new provider, create a new class derived from
```
andreww, line 6: We should probably create space in the documentation for how to do this. Happy to write it as I work out how (probably via wattnet) or you can put a stub in place as part of this PR.

cats/providers/uk_carbonintensity.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Any, ClassVar
+from zoneinfo import ZoneInfo
+
+from typing_extensions import override
+
+from ..exceptions import InvalidLocationError
+from ..forecast import PointEstimate, Timeseries
+from .base import BaseProvider, fetch_url, provider
+
+UK_POSTCODE_REGEX = re.compile(r"^[A-Z]{1,2}\d[A-Z\d]?")
```
andreww: Reminds me that I noticed we can do (much) better than this for postcode validation. Not in this PR (maybe it should be a totally separate package we import - in-fact uk-postcode-utils exists on PyPI...). It's been a while since I was good with regexes. I think this will accept valid postcodes fine but that some invalid postcodes will end up being accepted too.

andreww: (And feel free to ignore me here!)

cats/cli.py
```diff
@@ -288,55 +295,41 @@ def positive_integer(string):
...
-    try:
-        CI_forecast = get_CI_forecast(location, CI_API_interface)
-    except InvalidLocationError:
-        logging.error(f"Error: unknown location {location}\n")
-        logging.error(
-            "Location should be be specified as the outward code,\n"
-            "for example 'SW7' for postcode 'SW7 EAZ'.\n"
-        )
-        return 1
+    provider = provider_cls(location)
```
andreww: I'm not sure how I feel about having the location as an argument to the provider __init__ method and the start time as an argument to the .get_data method. It mirrors what we previously did but are these arguments fundamentally different? 

For what it's worth, I'm using CATS as a imported module into a very simple script where I download data for a bunch of different locations every 6 hrs with code at https://github.com/GreenScheduler/get_carbon_intensity_data. It's super simple but it is a different use of the API.

Not saying this is wrong, just don't know what is the best abstraction I guess.

tests/test_regressions.py
```diff
@@ -1,29 +1,51 @@
...
-    for idt, v in zip(range(19),
-                      [14, 20, 20, 22, 13, 40, 49, 76, 91, 86, 
-                       61, 28, 24, 26, 25, 27, 27, 26, 23])
+        PointEstimate(
+            datetime=d + timedelta(minutes=idt * 30),
+            value=v,
+        )
+        for idt, v in zip(
+            range(19),
+            [
```
andreww: Why?

andreww (commented):
Happy overall. Probably worth briefly thinking about the location / time aspect of the API. Does this also close some further issues (e.g. renaming CI modules)? If so worth making sure we capture these.

andreww:
The readthedocs and cluster-tests failures look like they linked. Both report not being able to import "cats.providers", but I cannot immediately reproduce locally.

cats/providers/uk_carbonintensity.py
abhidg: ONS has postcode data, could extract the first part into a text file, should be a few KB.

cats/cli.py
abhidg: Yeah, could move location to `get_data()` and keep `__init__()` only for authentication or empty for open APIs like carbonintensity.org.uk

tests/test_regressions.py
abhidg: This is a autoformat, can revert it

cats/providers/base.py
abhidg: added a method `get_temporal_resolution_minutes()`

andreww (approved):
Looks good to me now the tests are passing

andreww:
One thing to do when merging this is to open a documentation issue for information around how to add a new data provider. I'm happy to write the documentation as I knock up a Wattnet interface.

PR #199, opened 2026-07-29
Add grant codes to documentation

andreww:
A simple change to the documentation - note that we need this for future research output reporting.


abhidg (approved):
looks good

PR #200, opened 2026-07-30
Add wattnet.eu as a provider

### LINKED ISSUE #22: Multinational support: grid APIs for other countries

### LINKED ISSUE #44: Providing users with the option to pass their own API wrappers for carbon intensity
This is to continue the discussion @tlestang started with PR #43 

From my comments there: 

> I see two different use cases here:
> 1. We or other contributors will want to add other CI APIs (e.g. for other countries), and we ideally want to make them part of CATS so that these new APIs are available to the whole community. In this case, it would be good to have all the URL/parsing codes in the same place and api_inferface is a good place for that (it also makes it easier to add things by copy-pasting). And in terms of how much hassle it is to add it, it's equivalent now and with the new code (api_interface needs to be modified either way, and current code requires messing with __init__.py as well), but the existing code doesn't allow user to easily pick an API, this is what the new argument --api-carbonintensity introduces.
>
> 2. Second use case is if users want to pass their own API wrapper directly to CATS without having to modify the code. And in this case I agree, it would be good to make it possible in an easier way. But how would that work in practice? It would be good to have an idea of how the user would do it if we want to implement it.

This issue is to discuss whether we want to implement (2) and how it would work in practice from the user's point of view.

### LINKED ISSUE #124: Real time carbon intensity France
Hi,

As you mentioned: "Currently, CATS only works in the UK. If you are aware of APIs for real-time grid carbon intensity data in other countries, please open an issue and let us know."

For France, you can check out the [EcoWatt API](https://www.monecowatt.fr/faqs/comment-r%C3%A9cup%C3%A9rer-en-temps-r%C3%A9el-les-donn%C3%A9es-des-signaux-ecowatt/).

It provides a green, orange, or red signal depending on the electric system's load, which might serve as a good approximation of carbon intensity.

Best,
Thomas

### LINKED ISSUE #198: docs: information around how to add a new data provider
CATS has a new providers interface at `cats.providers`, documentation can work through an example provider showing how to integrate with existing CATS code.

andreww:
This needs work on error checking, properly integrating the auth stuff and plugging into the rest of CATS, but it is hopefully a useful start. Also needs tests! However, this allows us to run:

    import datetime
    import cats.providers.eu_wattnet
    t = datetime.datetime.now()
    C = cats.providers.eu_wattnet.WattnetEuProvider()
    C.update_authorization_token()
    C.get_data(timestamp=t, location="GB")

In order to get the auth to work you'll need an account to be set up and have your email address and password in environment variables:

    export CATS_WATTNET_EMAIL=<name>@<domain>
    export CATS_WATTNET_PASSWORD=<secret>

Not clear how best to have this running in testing. No quotes!


abhidg:
thanks @andreww, I have tried it and getting data works!

andreww:
Interesting to see the difference between wattnet.eu (first) and carbonintensity.org (second).

Figure_1

Figure_2

Cannot compare these directly (they cover different areas - all of the GB grid and just the "South England" distribution area). This will explain some of the smoothing I expect.

andreww:
Work for me to do here:

* Add error checking to our HTTP get function
* Sort out the typing errors
* Add tests for wattnet
* Add documentation on how to add a provider

andreww:
This PR is not going to deal with #201. That's for later.

andreww:
@abhidg  - any idea why that cluster test has started failing? I don't think I've done anything that should impact this but it looks like the SLURM queue isn't working as expected.

andreww:
Anyway, this seems to work. I've not added the general developer documentation on how to add a provider but otherwise this gives basic support for wattnet and scheduling across Europe.

cats/providers/base.py
```diff
@@ -26,21 +26,43 @@
...
+    Fetch and decode json representation of a provider's forecast data
+
+    The provider is responsible for formatting a `url` which typically includes
+    location, time and format information for the desired forecast. This is
+    either extracted from a cache (if the forecast has been requested before with
+    the same URL) or requested from the provider. The provider can also add `headers`
+    as needed. This function always includes the CATS user_agent in the headers
+    used in the request. Successful responses are cached, decoded from json to 
+    python objects and returned to the provider (which is responsible for extracting
+    the required information). Python objects may be returned as a dictionary, list,
```
abhidg: dictionary or list (those are the only two options)

cats/providers/eu_wattnet.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            patch_minute = 1
+        start_time = timestamp.replace(minute=patch_minute, second=0,
+                                       microsecond=0)
+        end_time = start_time + datetime.timedelta(
+            minutes=self.get_max_duration_minutes())
+        
+        # Build URL. Note that because we use timezone aware datetime object
+        # (and force them into UTC) .isoformat() adds +00:00 to the end of 
+        # the times in the URL. This breaks things. Instead we use strftime
```
abhidg, line 122: strftime is not used here

cats/providers/eu_wattnet.py
```diff
@@ -0,0 +1,182 @@
...
+            f"zone={location}&"
+            f"start={start_time.strftime('%Y-%m-%dT%H:%M:%S')}&"
+            f"end={end_time.strftime('%Y-%m-%dT%H:%M:%S')}"
+        )
+
+        # Setup authentication if this has not been done
+        if ((self.api_data is None) or
+            (self.api_data['expires_at'] < datetime.datetime.now())):
+            self.update_authorization_token()
+        assert self.api_data is not None, "Unexpected Wattnet authentication error"
```
abhidg: can be an exception

docs/source/quickstart.rst
```diff
@@ -42,6 +42,47 @@ should be transparent to cluster users.
 It will display the time to start the job on standard out and optionally
 some information about the carbon intensity on standard error.
 
+Locations outside Great Britain
```
abhidg, line 45: should we use Great Britain elsewhere in some places we refer to UK

docs/source/quickstart.rst
```diff
@@ -42,6 +42,47 @@ should be transparent to cluster users.
...
+-------------------------------
+
+The default behavior of CATS is to use carbon intensity forecast data provided
+by the National Energy System Operator (NESO), which manages the electricity grid in 
+Great Britain (Northern Ireland uses a separate grid operating across
+the ireland of Ireland). In order to use CATS for locations across Europe
```
abhidg: island

docs/source/quickstart.rst
```diff
@@ -42,6 +42,47 @@ should be transparent to cluster users.
...
+it is possible to specify a different experimental data source provided by
+the https://wattnet.eu/ project. This provides data at coarser spatial resolution
+(approximately 60 zones across Europe including one representing Great Britain
+compared to 14 areas provided by by the NESO) but at finer time resolution (15
```
abhidg: duplicate by

abhidg (approved):
Looks good @andreww a couple of minor points

abhidg:
For the failing test, begin time was set to before the submission time somehow which shouldn't happen and that triggered a failure since the testing script couldn't find the `BeginTime` flag.

docs/source/quickstart.rst
andreww: Yes. But I think we do. I've not found "UK" or "United" or "Kingdom" other than "uk" related to URLs.

cats/providers/eu_wattnet.py
andreww: We do, line 129 and 130, just not in the assert. Have clarified

